PR #15, opened 2020-07-18
Set up CI with Azure Pipelines

pingzing:
Sets up some basic CI in Azure DevOps. You can see [how it turned out here](https://dev.azure.com/mcalistern/offprint-fork/_build).

A few caveats:
 * Tests in both frontend and backend fail. Backend because they're just default tests, frontend because the test agent isn't running an X server, and can't run the tests. I left the tests enabled for now, because ¯\\\_(ツ)_/¯
* Part of the backend CI process fetches and runs MongoDB in a docker container, which isn't _strictly_ necessary, as it never actually runs the backend. But in the future, if any of the integration tests require a database, it's nice to have it there.
* `node_modules` is cached using [DevOps' caching feature](https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops). It saves about ~20s per build.
* I didn't even sketch out a deploy task, because I have nooooo idea what the deploy process would look like. But fundamentally, I'd probably divide the build process into two `stage`s, and publish artifacts from the "build" stage to a theoretical "deploy" stage, then send it off to Digital Ocean, or wherever.

### DIFF
--- a/.azure-pipelines/azure-pipelines-backend.yml
+++ b/.azure-pipelines/azure-pipelines-backend.yml
@@ -16,10 +16,9 @@ steps:
       versionSpec: '12.17.0'
     displayName: 'Install Node.js 12.17.0'
 
-  - script: |
-      yarn global add @angular/cli@8.3.21
-      yarn global add @nestjs/cli@6.12.6
-      yarn global add typescript@3.7.2 
+  - script: |      
+      yarn global add @nestjs/cli@7.1.5
+      yarn global add typescript@3.8.3
     displayName: 'Install global tools'
 
   - script: |
--- a/.azure-pipelines/azure-pipelines-backend.yml
+++ b/.azure-pipelines/azure-pipelines-backend.yml
@@ -1,0 +1,50 @@
+trigger:
+  branches:
+    include:
+      - master
+  paths:
+    exclude: 
+      - /frontend
+      - /.azure-pipelines/azure-pipelines-frontend.yml
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps: 
+  - task: NodeTool@0
+    inputs:
+      versionSpec: '12.17.0'
+    displayName: 'Install Node.js 12.17.0'
+
+  - script: |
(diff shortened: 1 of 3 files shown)


nisamson:
I think there's a hint that a Docker file might be good for the project overall

.azure-pipelines/azure-pipelines-backend.yml
```diff
@@ -0,0 +1,50 @@
...
+steps: 
+  - task: NodeTool@0
+    inputs:
+      versionSpec: '12.17.0'
+    displayName: 'Install Node.js 12.17.0'
+
+  - script: |
+      yarn global add @angular/cli@8.3.21
+      yarn global add @nestjs/cli@6.12.6
+      yarn global add typescript@3.7.2
```
Figments: Sorry about this! Check the README for updated versions of Angular, Nest, and TypeScript.

.azure-pipelines/azure-pipelines-frontend.yml
```diff
@@ -0,0 +1,48 @@
...
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '12.17.0'
+  displayName: 'Install Node.js 12.17.0'
+
+- script: |
+    yarn global add @angular/cli@8.3.21    
+    yarn global add typescript@3.7.2
```
Figments: Same goes for this one

.azure-pipelines/azure-pipelines-backend.yml
pingzing: Updated. Also removed Angular, as the backend doesn't use it.

.azure-pipelines/azure-pipelines-frontend.yml
pingzing: Updated.

pingzing:
That'll teach me to trust a readme instead of the actual code =P

Figments:
Alright, looks good! 

Merging.

PR #18, opened 2020-07-19
Added Dockerfile with docker-compose for simple dev deployment

nisamson:
Run `docker-compose up` to launch the the service. Set up to allow rapid rebuilds by mounting the source directory rather than baking it into the image.


Figments:
Looks good! Let's merge.

PR #19, opened 2020-07-19
Added Docker section to README; added sample.env for rapid env setup

nisamson:
Added a small set of instructions to get a dev environment up and running with only a few steps.


Figments:
Looks good! Merging now.

PR #20, opened 2020-07-19
Add native word counter module

### LINKED ISSUE #12: Add word counting functionality
This stuff should be done in the frontend in order to reduce server strain on the backend. The resulting word count should be sent along with other relevant content information to the backend, and inserted normally.

Voca.js will be the library used for this. The only problem is converting the [QuillJS Deltas](https://quilljs.com/docs/delta/) into a string of words that can actually be counted.

But that shouldn't be too difficult.

pingzing:
Things that should probably happen before this PR gets merged:

- [x] Figure out _how_ and _when_ to build native modules
    - Done. A topmost `Cargo.toml` that tracks all the native modules. Then we can just make `yarn` do `cargo build --release`.
- [x] Figure out the folder structure around native modules
     - Done. Each native module gets a folder in `/native`, and that folder also includes `.ts` file that wraps up the native import, etc.
- [x] Come up with a sane way of dealing with return values across FFI boundaries
    - No longer relevant!
- [x] Update the readme to address native stuff
    - Done, added a sub-readme inside `/native`.


native/rust/word_counter/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct DeltaOps {
+    attributes: Option<Value>,
+    insert: Value,
+}
+
+#[no_mangle]
+pub extern "C" fn count_words(work_content: *const c_char, ) -> u32 {
```
pingzing, line 27: If we wanted to change this signature to be capable of passing back an error message, it would become something like:
```rust
pub extern "C" fn count_words(work_content: *const c_char, result: *const c_char, result_length: u32) -> u32 {
...
```

src/db/works/works.service.ts
```diff
@@ -1,16 +1,24 @@
...
 export class WorksService {
+    
+    private wordCounterLib: any;
+
     constructor(
         @InjectModel('Work') private readonly workModel: Model<models.Work>,
         @InjectModel('Section') private readonly sectionModel: Model<models.Section>,
-        private readonly usersService: UsersService) {}
+        private readonly usersService: UsersService) {
+            this.wordCounterLib = new Library("../../native/compiled/word_counter", {
```
pingzing: This is another one of those things that should probably be centralized--I was thinking that we'd want all of our native modules to live in `/native/compiled/`, and soooome part of the build process generates them. That, or we build them by hand, and just leave that folder as something only seldom-touched.

native/rust/word_counter/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct DeltaOps {
+    attributes: Option<Value>,
+    insert: Value,
+}
+
+#[no_mangle]
+pub extern "C" fn count_words(work_content: *const c_char) -> u32 {
```
pingzing: If we wanted to change this signature to be capable of passing back an error message, it could become something like:
```rust
pub extern "C" fn count_words(work_content: *const c_char, result_message: *const c_char, result_length: u32) -> u32 {
...
```

nisamson:
I'd argue we should be relying on the system to figure out where the library is; that would just mean adding the shared object to `LD_LIBRARY_PATH`. That'd obviate the issue of deciding where it lives statically.

pingzing:
TIL about `LD_LIBRARY_PATH`. Neat.

Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[workspace]
+
+members = [
+    "native/word_counter",
```
pingzing, line 7: this is my solution to future native-modules: each one is a separate Rust crate

frontend/package.json
```diff
@@ -4,6 +4,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
+    "start-dev": "ng serve --proxy-config proxy.config.json",
```
pingzing: (so wait, how did you two have both backend and frontend running simultaneously without this)

native/compiled/export.ts
```diff
@@ -0,0 +1,5 @@
+const word_counter = require('./word_counter.node');
+
+export function count(bodyContent: string): number {
```
pingzing: no merging till docs and error-handling happen

Also, not really sure where this file should live? Next to the `.so` file? Next to native module's `Cargo.toml`? ¯\\\_(ツ)_/¯

native/copyNativeLibs.ts
```diff
@@ -0,0 +1,112 @@
+// Heavily inspired by napi-rs's "napi" script, (https://github.com/napi-rs/napi-rs/blob/master/scripts/napi.js)
```
pingzing: This entire goddamn file is a glorified copy/rename

native/word_counter/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "word_counter"
+version = "0.1.0"
+authors = ["pingzing <mcalistern@outlook.com>"]
```
pingzing: how do we want to handle authors, btw? or should I just axe this line entirely?

Figments: I'd axe that line, yeah. It's part of the Offprint project, so the "authors" are all contributors.

package.json
```diff
@@ -8,6 +8,7 @@
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build",
+    "buildNativeModules": "cargo build --release && ts-node ./native/copyNativeLibs.ts --release ./native/compiled/",
```
pingzing: this compiles everything it finds in our root-most Cargo.toml, then plucks the compiled `.dll/dylib.so`s out of the target folder, and stuffs them all into `/native/compiled`. Would love feedback on folder organization here

package.json
```diff
@@ -64,8 +66,11 @@
     "eslint-config-prettier": "^6.10.0",
     "eslint-plugin-import": "^2.20.1",
     "jest": "^25.1.0",
+    "minimist": "^1.2.5",
+    "napi-rs": "^0.2.4",
```
pingzing: probably don't need this, actually

native/word_counter/Cargo.toml
nisamson: If we set up a project email account, it'd be a good idea to put that in here.

Figments: 100% agree. I have one kicking around but I'll need to re-enable the Google Teams thing on my domain.

pingzing: axed :+1:

Dockerfile
```diff
@@ -8,12 +8,20 @@ RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources
...
+    libclang-dev \
+    clang
 
 RUN npm install -g typescript@3.8.3 \
     @angular/cli@9.1.5 \
     @nestjs/cli@7.1.5
 
+# Download rust and add it to the PATH
+RUN curl https://sh.rustup.rs -sSf |  bash -s -- -y
+RUN echo 'source $HOME/.cargo/env' >> $HOME/.bashrc
```
pingzing, line 23: Adds Rust, and all the stuff necessary to build with NAPI.

package.json
```diff
@@ -8,6 +8,8 @@
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build",
+    "buildNativeModules": "cargo build --release && ts-node ./scripts/copyNativeLibs.ts --release",
+    "postbuild": "mkdir ./dist/native/compiled && cp ./native/compiled/* ./dist/native/compiled/",
```
pingzing: Because Nest doesn't care about our `.node` files.

native/copyNativeLibs.ts
pingzing: Okay, it's a little better now. I tore out the stuff we don't use, and its defaults are ones we want.

package.json
pingzing: No longer necessary!

native/README.md
```diff
@@ -0,0 +1,156 @@
+# Native Modules 
+
+Some portions of the pulp-fiction backend are written in native code for the sake of performance. The source code for thoe native modules can be found here.
```
Figments, line 3: "The source code for **those**"

native/README.md
```diff
@@ -0,0 +1,156 @@
...
+Native modules are written in [Rust](https://www.rust-lang.org/), and use [NodeJS's N-API ](https://nodejs.org/api/n-api.html) to communicate with the rest of the backend.
+
+The basics of creating a new native module are the following:
+
+First, **from the root folder of the repository**, create a new home for your module:
+```bash
+cargo new ./native/your_module_name
+```
+
+Cargo will complain about the package beliieving it's in a workspace, when it's not. It is correct! We need to go add our new module to the topmost `Cargo.toml`, at the repository root.
```
Figments, line 25: "complain about the package **believing**"

pingzing:
Okay, this should be merge-worthy now. Things I settled on:
* `LD_LIBRARY_PATH` gud. We now have a helper that goes and reads the env var, and looks for a folder named `pulpfiction`, then reads from there.
* By default, our native modules get compiled, then shipped over to `/usr/local/lib/pulpfiction/`.

One thing that _doesn't_ get done right now: _nothing ever actually defines or updates `LD_LIBRARY_PATH` to add `/usr/local/lib/pulpfiction` to it_. This is because there are about three competing ways of doing so, and I'm not knowledgeable enough to make the call for which we should do.

Figments (approved):
Looking very good! Despite it being almost 2AM at time of posting, I can follow along relatively easily. 

Great job!

PR #26, opened 2020-07-23
Switching to Neon and GitHub Packages for all necessary modules

Figments:


Figments:
It's counting long strings correctly. Looks like this is the way to go from now on.

PR #28, opened 2020-07-23
Remove last traces of old native module approach

pingzing:
A few stray pieces that we don't need anymore.


Figments (approved):
Looks good.

PR #32, opened 2020-07-25
Expose debugging ports, add Neon CLI to Dockerfile

pingzing:


Figments (commented):
The last file says it should be removed, and since you already ran the benchmarks, is this still necessary to keep around?

pingzing:
Oops, missed that when doing cleanup. Yep, fixed now.

PR #33, opened 2020-07-29
Add invite code stuff

### LINKED ISSUE #22: Make sure invite codes are ready to go by launch
Invites must be manually created and added to the database.

pingzing:
Closes #22.

Argh, my autoformatter added spaces to all the inline object braces. Try to ignore the noise =/


frontend/src/app/models/users/create-user.model.ts
```diff
@@ -2,4 +2,5 @@ export interface CreateUser {
     readonly email: string;
     readonly username: string;
     readonly password: string;
+    readonly inviteCode?: string;
```
pingzing, line 5: Nullable so it's safe to remove later

yes I know MongoDB is schemaless _old habits die hard okay_

src/db/users/users.service.ts
```diff
@@ -1,38 +1,59 @@
 import { Injectable, ConflictException, BadRequestException, InternalServerErrorException } from '@nestjs/common';
 import { InjectModel } from '@nestjs/mongoose';
 import { Model } from 'mongoose';
-import { isNullOrUndefined } from 'util';
 import { hash, argon2id } from 'argon2';
 import * as sanitize from 'sanitize-html';
 import validator from 'validator';
-
 import * as models from './models';
+import { isNullOrUndefined } from '../../util/validation';
```
pingzing, line 9: The version bundled with Node is deprecated, apparently.

src/db/users/users.service.ts
```diff
@@ -1,38 +1,59 @@
...
-            const existingEmail = await this.userModel.findOne({email: sanitize(newUserInfo.email)});
-
-            if (isNullOrUndefined(existingUsername) && isNullOrUndefined(existingEmail)) {
-                const newUser = new this.userModel(newUserInfo);
-                return await newUser.save();         
-            } else {
-                throw new ConflictException('Someone already has your username or email. Try another combination.');
-            }
-        } else {
+        if (!newUserInfo.inviteCode) {
```
pingzing, line 27: In addition to the extra validation, I reworked this to use the early return pattern. I tend to prefer it in validation-heavy stuff, because then the success value is always at the very bottom of the function.

src/db/users/users.service.ts
```diff
@@ -151,11 +172,11 @@ export class UsersService {
...
      */
     async changeNameAndEmail(userId: string, newNameAndEmail: models.ChangeNameAndEmail): Promise<models.User> {
-        const existingUsername = await this.userModel.findOne({username: sanitize(newNameAndEmail.username)});
-        const existingEmail = await this.userModel.findOne({email: sanitize(newNameAndEmail.email)});
+        const existingUsername = await this.userModel.findOne({ username: sanitize(newNameAndEmail.username) });
+        const existingEmail = await this.userModel.findOne({ email: sanitize(newNameAndEmail.email) });
 
-        if (isNullOrUndefined(existingUsername) || isNullOrUndefined(existingEmail)) {
-            return await this.userModel.findOneAndUpdate({"_id": userId}, {"email": newNameAndEmail.email,"username": newNameAndEmail.username});
+        if (isNullOrUndefined(existingUsername) && isNullOrUndefined(existingEmail)) {
```
pingzing, line 181: This is a behavior change. Can you double-check to make sure it makes sense?

src/util/validation.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Checks whether a value is null or undefined. Replaces
+ * Node's util.isNullOrUndefined.
+ * @param value The value to check for null-or-undefinedness.
+ */
+export function isNullOrUndefined(value: any): boolean {
```
pingzing, line 6: whee, one line functions. thanks Node, I love it

src/db/users/users.service.ts
Figments: Originally it was set to `&&` but the problem is that it means *both* the username and email cannot already exist, meaning if a user is only changing their username and NOT their email, this wouldn't work, as it's supposed to replace both of them in one go (using any pre-existing values accordingly).

We should probably investigate that at a later date, because emails/usernames should be unique, but the method by which either one is updated (or both for that matter) should probably be done different to avoid these situations.

Figments (approved):
Alrighty, everything looks good to go!

PR #34, opened 2020-07-30
Added search functions

nisamson:
I think this is about where I'm going to have to leave things for now; I've got a lot of work stuff coming up. The search functionality should be correct, but it's not hooked up to the API.

In the future, we really need to figure out caching for searches; as is, this is going to be pretty expensive computationally.


Figments (approved):
Everything looks good. Merging.

PR #38, opened 2020-08-01
Add image uploading service

pingzing:


Dockerfile
```diff
@@ -28,12 +28,7 @@ RUN npm install -g typescript@3.9.7 \
 RUN curl https://sh.rustup.rs -sSf |  bash -s -- -y
 RUN echo 'source $HOME/.cargo/env' >> $HOME/.bashrc
 
-# Install all build dependencies
-RUN ./build_dev.sh
```
pingzing, line 32: Removed this from the Dockerfile because the contents of the current directory aren't visible to the container (and thus the Dockerfile) until _after_ this is done running.

Removed the start command because we haven't done `yarn install` yet. 

I think we need Cam's help here >.>

src/api/images/images.service.ts
```diff
@@ -0,0 +1,72 @@
...
+            Key: `${userId}-${uuidV4()}.${extension}`, // Note: Keys must be unique per-Space
+            Body: imageFile.buffer,
+            ACL: 'public-read',
+            ContentType: imageFile.mimetype,
+        };
+
+        return new Promise((resolve) => {
+            this.s3.upload(uploadOptions, null, (err: Error, data: AWS.S3.ManagedUpload.SendData) => {
+                if (err) {
+                    throw new InternalServerErrorException(err);
```
pingzing, line 42: This _will_ throw out to callers if they've awaited this method, right?

Figments: Yup. Any time a NestJS exception is thrown, no matter where it is, the function it's in will error out.

Figments (approved):
Everything looks good. I'll open an issue for Cam to take a look at re: Dockerfile.

PR #44, opened 2020-08-05
Add auto-refresh on 401

pingzing:


frontend/src/app/services/auth/auth.guard.ts
```diff
@@ -27,7 +27,7 @@ export class AuthGuard implements CanActivate, CanActivateChild {
     if (currentUser && currentUser.token) {
       if (this.helper.isTokenExpired(currentUser.token)) {
         this.authService.refreshToken().pipe(first()).subscribe(isValid => {
-          return isValid;
+          return isValid !== null;
```
pingzing, line 30: The reason we need to double-click on things when our refresh token expires is because this observable has no subscribers. This `canActivate()` should probably return an `Observable<boolean>`. https://stackoverflow.com/a/50143721/1639835 seems to have a good idea here.

frontend/src/app/services/auth/auth.guard.ts
```diff
@@ -53,7 +53,7 @@ export class AuthGuard implements CanActivate, CanActivateChild {
...
```
pingzing, line 56: ditto above

Figments (approved):
Looks good! I'll take a look at the frontend auth guard to see what's up.

PR #45, opened 2020-08-06
Add favicon, improve Beatriz hero image, improve theme handling

pingzing:


frontend/src/app/components/beatriz-hero/beatriz-hero.component.less
```diff
@@ -0,0 +1,9 @@
+.outline {
```
pingzing, line 1: These get used in the SVG

Figments (approved):
All looks good!

PR #49, opened 2020-08-08
Make auth excellent

### LINKED ISSUE #41: Add additional details to audit sessions
AKA refresh tokens. Recording the gist of our talk in Discord for posterity:

Right now, audit sessions live forever. Ideally, an audit session should be
- the session ID
- the expiration date
- what device the user signed in to
- when the session was created

Which would allow cleaning them out after they've expired, and (eventually?) allowing a user to force-logout a session.

pingzing:
Closes #41.

- Audit Sessions are now fully-fledged objects, including creation and expiration dates
- Refresh tokens are now SHA256-hashed in the database
- JWT configuration options are centralized in /src/util/secrets.ts
- Failing to refresh your token now results in a 403
- Receiving a 403 on refresh attempt will now log a user out

### DIFF
--- a/src/db/users/audit-session.schema.ts
+++ b/src/db/users/audit-session.schema.ts
@@ -1,0 +1,9 @@
+import { Schema } from "mongoose";
+
+import { AuditSession } from './models/audit-session.model';
+
+export const AuditSessionSchema = new Schema<AuditSession>({
+    _id: {type: String, required: [true, `You must provide a sessionId.`]},
+    expires: {type: Date, required: [true, `You must provide an expiration date.`]},
+    createdAt: {type: Date, default: Date.now()},
+});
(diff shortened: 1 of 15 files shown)


src/db/users/audit-session.schema.ts
```diff
@@ -0,0 +1,9 @@
+import { Schema } from "mongoose";
+
+import { AuditSession } from './models/audit-session.model';
+
+export const AuditSessionSchema = new Schema<AuditSession>({
+    _id: {type: String, required: [true, `You must provide a sessionId.`]},
+    expires: {type: Date, required: [true, `You must provide an expiration date.`]},
+    createdAt: {type: Date, default: Date.now()},
+});
```
pingzing, line 9: I tried to set up pre-save hooks to do the hashing and creation dates the easy way, but it seems like they don't fire when adding a sub-document to a parent document.

...or something.

Figments (approved):

PR #53, opened 2020-08-09
Add Browse page to frontend

pingzing:
exactly what it say on the tin

image


pingzing:
and now with better-looking metadata: 
image

PR #68, opened 2020-08-10
Give cover art uploader stronger opinions

pingzing:
Everything is JPEGs.

This seems to be the most reliable way to _not_ generate images that are larger than the original when cropping to its original size.


Figments (approved):

frontend/src/app/components/modals/works/upload-coverart/upload-coverart.component.ts
```diff
@@ -92,7 +89,8 @@ export class UploadCoverartComponent implements OnInit {
     }
 
     this.croppedImage = event.base64;
-    this.fileToReturn = this.base64ToFile(event.base64, 'coverart');    
+    this.fileToReturn = this.base64ToFile(event.base64, 'coverart');   
+    console.log(`Cropped file is ${((event.base64.length * 6) / 8).toLocaleString()} bytes.`);
```
pingzing, line 93: I left this log in on purpose--if people are still having issues, we can just ask them "what does your console say"

PR #74, opened 2020-08-11
Tweak Browse page slightly to make it better handle long titles

pingzing:
Old:
image

New: 
image

### DIFF
--- a/frontend/src/app/pages/browse/browse.component.html
+++ b/frontend/src/app/pages/browse/browse.component.html
@@ -13,14 +13,14 @@
                 <div style="margin: 0 auto 1rem auto;" *ngFor="let work of works">
                     <div class="work-container">
                         <!--Header-->
-                        <div class="work-header">
-                            <div style="display: inline-block;" [ngSwitch]="work.meta.rating">
+                        <div class="work-header">                            
+                            <ng-container [ngSwitch]="work.meta.rating">
                                 <img class="rating" src="assets/icons/teenIcon.svg" *ngSwitchCase="'Teen'">
                                 <img class="rating" src="assets/icons/matureIcon.svg" *ngSwitchCase="'Mature'">
                                 <img class="rating" src="assets/icons/explicitIcon.svg" *ngSwitchCase="'Explicit'">
                                 <img class="rating" src="assets/icons/everyoneIcon.svg" *ngSwitchDefault>
-                            </div>
-                            <h1><a [routerLink]="['/work', work._id, work.title | slugify]">{{ work.title }}</a></h1>
+                            </ng-container>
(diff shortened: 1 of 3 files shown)


frontend/src/app/pages/browse/browse.component.html
```diff
@@ -13,14 +13,14 @@ <h2>Newly Added</h2>
                 <div style="margin: 0 auto 1rem auto;" *ngFor="let work of works">
                     <div class="work-container">
                         <!--Header-->
-                        <div class="work-header">
-                            <div style="display: inline-block;" [ngSwitch]="work.meta.rating">
+                        <div class="work-header">                            
+                            <ng-container [ngSwitch]="work.meta.rating">
```
pingzing, line 17: +1 to `<ng-container>`. It doesn't show up in the DOM, but is targetable by Angular directives.

frontend/src/app/pages/browse/browse.component.less
```diff
@@ -39,38 +39,40 @@ div.browse-container {
     
             div.work-header {
                 margin-bottom: 0;
+                   
                 img.rating {
-                    position: relative;
+                    float: left;
```
pingzing, line 44: Using the ol' box model here so we can get the short description wrapping around the cover art.

frontend/src/app/pages/browse/browse.component.less
```diff
@@ -39,38 +39,40 @@ div.browse-container {
...
                     width: 24px;
                     height: 24px;
                     border: none;
                     box-shadow: none;
                     top: 0.25rem;
                 }
-                h1 {
-                    display: inline-block;
+                h1 {     
+                    display: inline;
```
pingzing, line 52: Making the title `inline` rather than `inline-block` makes it so that the rating image will never occupy a line all by its lonesome.

frontend/src/app/pages/browse/browse.component.less
```diff
@@ -71,6 +72,7 @@ div.browse-container {
                     a {
                         color: whitesmoke;
                         padding: 5px 5px 2px 2px;
+                        word-break: break-all;
```
pingzing, line 75: Our worst-case scenario here is the user with 32 capital Ws for a name. Since the author byline is `inline-block` it won't wrap unless it _really_ has to, but if it does, it'll break wherever it can.

frontend/src/app/pages/browse/browse.component.less
```diff
@@ -92,14 +94,15 @@ div.browse-container {
...
             }
     
-            div.short-description {
-                display: flex;
+            div.short-description {                
                 flex-grow: 1;
                 color: whitesmoke;
                 font-size: 14px;
                 margin-top: 1.5rem;
+                margin-bottom: 1.5rem;
```
pingzing, line 102: Just for symmetry.

Figments (approved):

PR #91, opened 2020-08-12
Feature/fix change user identity

### LINKED ISSUE #36: Investigate reworking the changeNameAndEmail function
Per discussion in https://github.com/OffprintStudios/pulp-fiction/pull/33#discussion_r462598288, adding it here so it gets remembered for some time after launch

Would it make more sense to just split this out into two separate functionalities? a `changeEmail` and `changeUsername`, for example?

pingzing:
Closes #36.

NOTE: Changing usernames has been removed based on my discussion with RBDash in Discord. The hooks are all still there, there's just no way to call it. We should address in the future.

Few things:
 - Introduce a new `network-input` component. It renders a pair of input boxes: one up top, that has an edit button, and a password box just below it. It takes as input:
  - An `inputType` for the top box
  - A `placeholder` for the top box
  - A `class` string for the top box
  - An input value for the top box
  - A `class` value for the password box
  - An upload function that is fired when the user accepts the changes.

Caveats:
 - The visuals are super basic ATM--the password box should _probably_ be more connected to the top box.
 - I worry that the "submit" button isn't noticeable enough. Ideas?
 - It has no validation, and currently no way to pass in validators, because I have no idea how Angular validators even work. However we do it, they'd have to be something the user can pass in though.
  - Is this too hard to use for mobile and screen readers?


pingzing:
Little preview 
evenbetteremailstuff

frontend/src/app/pages/home/settings/settings.component.ts
```diff
@@ -117,31 +114,27 @@ export class SettingsComponent implements OnInit {
...
+        if (x) {
+          this.alertsService.success("Email successfully changed!");
+          return x.email;          
+        } else {
+          return null;
+        }
+      }), catchError(_ => {
+        return of(null);
+      }));
+  };
```
pingzing, line 137: The only "gotcha" bit about the `uploadFunction` that the `network-input` expects is that it expects a string on success, and interprets any falsy value as failure.

Figments (approved):

PR #92, opened 2020-08-13
Feature/shared models

pingzing:
Managed to de-duplicate a few copied interfaces, too!

A few very small caveats:

 - `Genres` is now defined as the intersection between `GenresFiction` and `GenresPoetry`, i.e.
```typescript
type Genres = GenresPoetry | GenresFiction;
```
 - Any models which were used for DB documents and used `string | someType` as its type to make Mongoose cooperate are now just `someType`. To make Mongoose cooperate, we can just use `//@ts-ignore` to turn off the type checker for that one line where we use the query. Not great, but a lesser evil than polluting the type signature, IMO.
- All the shared types are stuffed into a single `shared-types.ts` file for the Frontend. We could probably divide this up a bit more, but I was tired of folder-wrangling.


shared/models/collections/collection.model.ts
```diff
@@ -18,7 +16,7 @@ export interface Collection extends Document {
 /* Helpers for building queries */
 export interface WorkInfo {
     readonly _id: string;
-    readonly author: string | AuthorInfo;
```
pingzing, line 21: This is one example of the Mongoose thing I was talking about. Now if we try to a do a query on "author:" "somevalue", it'll complain. It's not actually a problem, it's just that the types library can't seem to model this correctly. The solution is just turning off the type checker for that one line.

src/db/works/works.service.ts
```diff
@@ -182,6 +186,7 @@ export class WorksService {
      * @param sectionId The section ID
      */
     async getSectionForUser(user: any, workId: string, sectionId: string): Promise<models.Section> {
+        //@ts-ignore
         const thisWork = await this.workModel.findOne({ "_id": workId, "author": user.sub, "sections": sectionId }).where("audit.isDeleted", false);
```
pingzing, line 190: ...like this.

PR #130, opened 2020-08-26
Bring word_counter module into the repo via wasm_bindgen

pingzing:
And a host of other small docker-y and package.json-y changes. I'll go through and comment the relevant ones...

...and a screenshot of the directory structure around the new Rust module stuff. `/word_counter/` is generated by `nx`'s `generate library` schematic.  Everything inside `word_counter/src/lib` is generated by the the WASM plugins we're using, and totally unmodified. `index.ts` is the only file that needs to be modified, and even then only lightly.
image


.gitignore
```diff
@@ -20,6 +20,7 @@ target
 
 # IDE - VSCode
 .vscode/*
+.devcontainer/*
```
pingzing, line 23: VSCode's Docker stuff.

.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
+{
```
pingzing, line 1: What I use to debug the backend. The port number doesn't work now that `nx` assigns it randomly, but if we can make it static again, it should work just fine

Dockerfile
```diff
@@ -19,16 +19,21 @@ RUN apt-get update && apt-get install -y \
     libclang-dev \
     clang
 
-RUN npm install -g typescript@3.9.7 \
+RUN yarn add --dev typescript@3.9.7 \
```
pingzing, line 22: No need to use NPM here. And no need for these to be global, either.
I'm not 100% certain if Typescript needs to be `--dev` or not? It worked locally, anyway.

README.md
```diff
@@ -12,20 +12,21 @@ It's still very much a work in progress.
...
-* NestJS 7.4.1
 * MongoDB 4.2.1
-* [Neon](https://neon-bindings.com/docs/getting-started) and the [Rust 1.15 (or later)toolchain](https://rustup.rs/).
-    * (Optionally) If not on Ubuntu 16.10 or later, you will also need [LLVM](https://releases.llvm.org/download.html), as native module compilation relies on LLVM's `libclang`. You will also need to set the `LIBCLANG_PATH` environment variable to LLVM's `/bin` directory.
+* The [Rust 1.15 (or later) toolchain](https://rustup.rs/).
+    * You may also need the following packages in order to compile the Rust packages:
+        * `build-essential`
+        * `llvm-dev`
+        * `libclang-dev`
+        * `clang`
```
pingzing, line 25: Note: We may not actually need these anymore! I need to rebuild the docker container without them and see if the Rust modules still build.

packages/client/src/app/app.module.ts
```diff
@@ -126,7 +126,8 @@ const toolbarOptions = [
     }),
   ],
   providers: [
-    {provide: HTTP_INTERCEPTORS, useClass: AuthInterceptor, multi: true},
+    {provide: HTTP_INTERCEPTORS, useClass: AuthInterceptor, multi: true,},
+    SlugifyPipe,
```
pingzing, line 130: 'cause we use it in code in the New Section page

packages/server/webpack.config.js
```diff
@@ -0,0 +1,19 @@
...
+        plugins: [
+            ...config.plugins,
+            new WasmPackPlugin({
+                crateDirectory: path.resolve(__dirname, "..", "..", "native", "word_counter", "src"),                
+                extraArgs: "--target bundler",
+                outDir: path.resolve(__dirname, "..", "..", "packages", "word_counter", "src", "lib"),
+                outName: "word_counter",
+                // Release should probably be in a separate webpack.config.prod.js or something. But for now...
+                forceMode: "release"
+            })
```
pingzing, line 16: So, I haven't tested it, but I _think_ we can just add a `new WasmPackPlugin({})` for each Rust module we want to use. The args I'm using here are super rough (see ` path.resolve(__dirname, "..", ".." etc` for one), and I think realistically, we'll want two webpack files, one for debug and one for prod. But this is a basic proof-of-concept!

We can also do any Webpack customization we want to the backend in here, so that's a nice lever to have available.

packages/word_counter/src/index.ts
```diff
@@ -0,0 +1,14 @@
...
+let word_counter_promise = import('./lib/word_counter');
+
+/** Counts the number of words in the given JSON blob full of Quill deltas.
+* @param {string} bodyText A JSON blob made of Quill Deltas.
+* @returns {number} The number of words in the submitted text.
+*/
+export async function countWords(bodyText: string): Promise<number> {
+    let wasm_module = await word_counter_promise;
+    return wasm_module.count_words(bodyText);
+}
```
pingzing, line 13: This is the "minimal boilerplate" I meant. Basically, we have to wrap the WASM module up in a little bit of asncy loading because Webpack 4 _cannot_ synchronously load a WASM module--it can only do it asynchronously. So, we import, wrap the async-ness up, and export something a little cleaner to use.

Figments (approved):

PR #140, opened 2020-08-30
Use new Froala editor everywhere possible

pingzing:
Major changes:
 - All DB documents that have a body that was formerly stored in Quill Deltas now have a `usesFroala` boolean that determines whether or not their bodies are stored as Deltas or HTML. We can query the DB to see if anything is still `false` to see how the migration is going.
 - All the `new-etc` pages exclusively use the Froala editor and send up HTML. 
 - All the `edit-etc` pages set `usesFroala` on their respective work/blog/doc to `true`. If their current content is in Quill Deltas, it'll first parse out the rendered HTML, and send that up instead of the usual Deltas. 
 - All the viewing pages _should_ check the `usesFroala` flag to determine whether to use the quill viewer/editor, and will swap between them as needed.
 - All backend HTML sanitation is now handled by the Rust Ammonia crate.
 - Added a new `count_plaintext_words` to the `count_words` module, so we can count words of our new HTML-only formatted bodies.


Dockerfile
```diff
@@ -16,11 +16,9 @@ RUN apt-get update && apt-get install -y \
     yarn=1.22.1-1 \
     build-essential
 
-RUN yarn add --dev typescript@3.9.7 \
-    @angular/cli@10.0.4 \
+RUN yarn global add @angular/cli@10.0.4 \
```
pingzing, line 19: Adding the various CLIs locally-only seems to just plain not work. Global it is.

`wasm-pack` and `typescript` on the other hand, we can get away with as devDependencies.

native/html_sanitizer/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+    .add_tag_attributes("div", &["style"])
+    .add_tag_attributes("p", &["style"])
+    .add_tag_attributes("a", &["style"])
+    .add_tag_attributes("h1", &["style"])
+    .add_tag_attributes("h2", &["style"])
+    .add_tag_attributes("h3", &["style"])
+    .add_tag_attributes("h4", &["style"])
+    .add_tag_attributes("h5", &["style"])
+    .add_tag_attributes("h6", &["style"])
+    .add_tag_attributes("img", &["style"])
```
pingzing, line 23: We can probably customize this a lot more. I predict we'll need to, in order to support the way Froala styles inline images.

native/word_counter/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "word_counter"
-version = "0.1.0"
-authors = ["Alyx Mote <figments@outlook.com>"]
+version = "0.2.0"
+authors = ["Offprint Development Team"]
```
pingzing, line 4: Yay? Nay?

packages/client/src/app/components/froala-editor/froala-editor.component.ts
```diff
@@ -0,0 +1,90 @@
...
+        align: 'left',
+        buttonsVisible: 3
+      },
+      moreRich: {
+        buttons: ['insertHR', 'insertLink', 'insertImage', 'quote', 'clearFormatting'],
+        align: 'left',
+        buttonsVisible: 3,
+      },
+      moreMisc: {
+        buttons: ['undo', 'redo', 'fullscreen', 'html'],
```
pingzing, line 60: Warning: the `html` button allows users to toggle between the HTML and the WYSIWYG view. Dunno if we want that, but I turned it on for now, because it's probably better on than off while we're in Alpha.

Figments (approved):

PR #148, opened 2020-09-02
Feature/new editor

Figments:
We're switching to CKEditor 5


packages/client/src/app/components/new-editor/new-editor.component.ts
```diff
@@ -0,0 +1,67 @@
...
+          items: [ 'insertimageviaurl' ]
+        }
+      }
+    }
+  };
+
+  pasteBeforeCleanup(pastedHtml: string): string {
+    // Strip out spurious newline characters. <br> or bust, baby
+    return pastedHtml.replace(/(?:\r\n|\r|\n)/g, '');
+  }
```
pingzing, line 49: We can get rid of this, it looks like CKEditor does it on its own.

pingzing (changes requested):

packages/client/src/app/components/new-editor/new-editor.component.ts
Figments: got it! will change it in master

PR #153, opened 2020-09-03
Make deployments work again

pingzing:
The only "big" change is that this makes the backend go back to serving static files out of the `static` folder underneath it. The deployment process puts the frontend in the `static` folder, and bob's yer uncle. More comments inline on individual files.


.azure-pipelines/azure-pipelines-backend.yml
```diff
@@ -3,43 +3,95 @@ trigger:
...
+  # This is included in package.json, but only under devDependencies, so this (production!) build won't see it.
+  # As a workaround, install it globally
+  - script: yarn global add wasm-pack
+    displayName: 'Install wasm-pack'
+
+  - task: Cache@2
+    inputs:
+      # Cache the tarred node_modules.xz file.
+      # Note that the cache key is the hash of of yarn.lock. If the lockfile changes, the cache gets invalidated.
+      key: 'v2 | yarn | "$(Agent.OS)" | $(Build.SourcesDirectory)/yarn.lock'
```
pingzing, line 29: Had to version the key, or it would keep looking for the old node_modules cache, which was a folder containing, well, node modules. We don't want to cache that! It takes forever to publish it, and it _also_ takes forever to tar-then-publish it. So instead, the thing that we cache is a pre-tarred node_modules.

.azure-pipelines/azure-pipelines-backend.yml
```diff
@@ -3,43 +3,95 @@ trigger:
...
+      path: '$(Build.SourcesDirectory)/tarred_node_modules'
+      cacheHitVar: 'NodeCacheRestored'
+    displayName: 'Cache node_modules'
+
+  # Untar the cached node_modules folder if we had a cache hit
+  - script: tar -xf $(Build.SourcesDirectory)/tarred_node_modules/node_modules.tar.xz --directory=$(Build.SourcesDirectory) --verbose
+    condition: eq(variables['NodeCacheRestored'], 'true')
+    displayName: 'Untar node_modules.xz'
```
pingzing, line 37: ...and if it's in the cache, we just pull it out and untar it. Which is much faster than getting the folder out of cache, then tarring it for publishing at the end!

.azure-pipelines/azure-pipelines-backend.yml
```diff
@@ -3,43 +3,95 @@ trigger:
...
+    displayName: 'Cache word_counter output'
 
-    # Install production-only packages if the cache doesn't exist, or is out of date.
+  # Cache html_sanitizer's output
+  - task: Cache@2
+    inputs:
+      key: 'cargo | "$(Agent.OS)" | $(Build.SourcesDirectory)/native/html_sanitizer/Cargo.lock'
+      path: '$(Build.SourcesDirectory)/native/html_sanitizer/target'
+      cacheHitVar: 'WordCounterCacheRestored'
+    displayName: 'Cache html_sanitizer output'
```
pingzing, line 53: This is gonna get repetitive as we add more native modules. We can probably split it out into a sub-task, and just pass in an array of module names. I'm pretty sure we can do that with Az DevOps.

package.json
```diff
@@ -6,6 +6,8 @@
     "nx": "nx",
     "start": "nx serve",
     "build": "nx build",
+    "build:ci:server": "nx build server --skip-nx-cache --prod", 
+    "build:ci:client": "nx build client --skip-nx-cache --prod",
```
pingzing, line 10: Added these two, because `nx build` liked to always try to build the client when I passed `--skip-nx-cache` to it. Decided to just make it explicit.

And we _might_ be able to use the nx cache in the future, but after looking at the docs, it's complicated enough that I dunno if we want to bother.

pm2.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+    apps: [{
+        name: "pulpd",
+        script: "/opt/pulpd/site/main.js",
+        cwd: "/opt/pulpd/site/",
+        env_production: {
+            "NODE_ENV": "production"
+        },
```
pingzing, line 8: This is a super rough implementation of what this file might look like, but it basically controls the runtime environment of the Pulp Fiction backend as managed by `pm2`. I can imagine we'd want to do some string injection as part of the deployment process, instead of using a hardcoded `/opt/pulpd/site/`, for example, or maybe add extra env vars.

But for now! Good enough.

Figments (approved):

PR #162, opened 2020-09-15
Feature/content filters

### LINKED ISSUE #114: Filter works by content rating
Allow users to filter out Mature and Explicit works separately from each other. Could probably do that by checking and setting cookies, so users can set the filters on a per-session basis.

Figments:
Closes #114


packages/client/src/app/app.module.ts
```diff
@@ -9,13 +9,13 @@ import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
 import { AuthInterceptor } from './services/auth';
 import { MaterialModule } from '@pulp-fiction/material';
-import { ToppyModule } from 'toppy';
```
pingzing, line 12: good riddance

packages/client/src/app/app.module.ts
```diff
@@ -105,7 +105,7 @@ const toolbarOptions = [
     NotificationsComponent, WatchingPageComponent, HistoryComponent, StartConversationComponent, TruncatePipe
   ],
   imports: [
-    BrowserModule, AppRoutingModule, HttpClientModule, FormsModule, ReactiveFormsModule, ToppyModule,
+    BrowserModule, AppRoutingModule, HttpClientModule, FormsModule, ReactiveFormsModule, CookieModule.forRoot(),
```
pingzing: should the CookieModule get a line of its own, to match the pattern we have down below (i.e. anything that does more than just get imported gets its own line)

packages/client/src/app/pages/home/settings/settings.component.html
```diff
@@ -68,5 +68,21 @@ <h1>Update Tagline</h1>
                 </div>
             </form>
         </div>
+        <div class="settings-card">
+            <form [formGroup]="setContentFilter" (ngSubmit)="submitContentFilter()">
+                <div class="settings-card-content" style="margin-bottom: 1.5rem;">
+                    <h1>Content Filters</h1>
+                    <span style="display: block; width: 100%; text-align: center;">
```
pingzing: why inline styles instead of adding those rules to the CSS classes?

packages/client/src/app/pages/home/settings/settings.component.html
```diff
@@ -68,5 +68,21 @@ <h1>Update Tagline</h1>
...
+                        <mat-slide-toggle formControlName="enableMature">See Mature</mat-slide-toggle>
+                    </span>
+                    <span style="display: block; width: 100%; text-align: center;">
```
pingzing: ditto here

packages/client/src/app/pages/home/settings/settings.component.ts
```diff
@@ -194,4 +237,8 @@ export class SettingsComponent implements OnInit {
       location.reload();
     });
   }
+
+  submitContentFilter() {
+      this.authService.setContentFilter(this.setFilterFields.enableMature.value, this.setFilterFields.enableExplicit.value);
```
pingzing, line 242: Am I correct in assuming content filtering is set _both_ locally in a `contentFilter` cookie, _and_ server-side? (and when a user logs in, their `contentFilter` cookie gets populated?)

packages/client/src/app/services/auth/auth.service.ts
```diff
@@ -108,7 +111,7 @@ export class AuthService {
     // Fire and forget. If this fails, it doesn't matter to the user, 
     // and we don't want to leak that fact anyway.
     this.http.get(`${this.url}/logout`, { withCredentials: true }).subscribe();
-
+    this.cookies.remove('contentFilter');
```
pingzing, line 115: nitpick: maybe leave a newline here, to make it clear the above comment only refers to the HTTP logout call

packages/client/src/app/services/auth/auth.service.ts
```diff
@@ -250,6 +253,26 @@ export class AuthService {
...
+      this.cookies.put('contentFilter', ContentFilter.MatureEnabled);
+    } else if (enableMature === false && enableExplicit === true) {
+      this.cookies.put('contentFilter', ContentFilter.ExplicitEnabled);
+    } else if (enableMature === true && enableExplicit === true) {
+      this.cookies.put('contentFilter', ContentFilter.Everything);
+    } else if (enableMature === false && enableExplicit === false) {
+      this.cookies.put('contentFilter', ContentFilter.Default);
+    }
+
+    location.reload();
```
pingzing, line 274: huh, do we really need to force a refresh before cookies get recognized?

packages/server/src/app/api/browse/browse.controller.ts
```diff
@@ -1,12 +1,15 @@
...
 import { BrowseService } from './browse.service';
 
 @Controller()
 export class BrowseController {
     constructor(private readonly browseService: BrowseService) {}
 
     @Get('all-pub-works/:pageNum') 
-    async getAllPutWorks(@Param('pageNum') pageNum: number) {
-        return await this.browseService.getAllNewPublishedWorks(pageNum);
+    async getAllPutWorks(@Cookies('contentFilter') contentFilter: ContentFilter, @Param('pageNum') pageNum: number) {
```
pingzing: this function name is misspelled. Should be `getAllPubWorks`

packages/server/src/app/db/works/works.service.ts
```diff
@@ -387,25 +387,99 @@ export class WorksService {
...
      * Fetches all new published works by newest first.
      */
-    async fetchNewPublishedWorks(pageNum: number): Promise<PaginateResult<documents.WorkDocument>> {
-        return await this.workModel.paginate({'audit.published': models.ApprovalStatus.Approved, 'audit.isDeleted': false}, {
-            sort: {'audit.publishedOn': -1},
-            page: pageNum,
-            limit: 15
-        });
+    async fetchNewPublishedWorks(contentFilter: models.ContentFilter, pageNum: number): Promise<PaginateResult<documents.WorkDocument>> {
+        if (contentFilter === models.ContentFilter.Everything) {
```
pingzing: There's a lot of duplication here. Could you do something like

```typescript
let query = {
  'audit.published': models.ApprovalStatus.Approved,
  'audit.isDeleted': false,
};
let queryOptions = {
  sort: {'audit.publishedOn': -1},
  page: pageNum,
  limit: 15
};
if (/* various different content filter whatever*/) {
  query['$or'] = // whatever the query is for this if-branch
}
else if (/*etc etc*/) {
  query['$or'] = // you get the idea
}
return await this.workModel.paginate(query, queryOptions);
```

packages/server/src/app/db/works/works.service.ts
```diff
@@ -387,25 +387,99 @@ export class WorksService {
...
      * @param userId The user whose works we're fetching
      */
-    async getWorksList(userId: any, pageNum: number): Promise<PaginateResult<documents.WorkDocument>> {
-        return await this.workModel.paginate({'author': userId, 'audit.published': models.ApprovalStatus.Approved, 'audit.isDeleted': false}, {
-            sort: {'audit.publishedOn': -1},
-            page: pageNum,
-            limit: 15
-        });
+    async getWorksList(userId: any, contentFilter: models.ContentFilter, pageNum: number): Promise<PaginateResult<documents.WorkDocument>> {
+        if (contentFilter === models.ContentFilter.Everything) {
```
pingzing: Same here as above

pingzing (changes requested):

packages/client/src/app/pages/home/settings/settings.component.ts
pingzing: ah, no, it's _only_ local. Okay. May want to store serverside in a later update

packages/client/src/app/app.module.ts
Figments: lmao

packages/client/src/app/app.module.ts
Figments: it should!

packages/client/src/app/pages/home/settings/settings.component.ts
Figments: something to consider, yeah.

the idea behind it only being stored locally was so users can have different filter settings depending on what browser/computer logged into. like, say, a user might want to have mature stories turned on at home but when they're just browsing at work they want that turned off.

packages/client/src/app/services/auth/auth.service.ts
Figments: yeah, without a refresh the cookie doesn't get stored for some reason.

packages/server/src/app/db/works/works.service.ts
```diff
@@ -388,46 +388,36 @@ export class WorksService {
...
-                page: pageNum,
-                limit: 15
-            });
+        let query = {'audit.published': models.ApprovalStatus.Approved, 'audit.isDeleted': false};
+        let paginateOptions = {sort: {'audit.publishedOn': -1}, page: pageNum, limit: 15};
+
+        switch (contentFilter) {
+            case models.ContentFilter.Everything: 
+                query = query;
+                break;
```
pingzing, line 397: if you wanted, you could even remove this arm entirely

pingzing (approved):

pingzing:
Looks good 👍

packages/server/src/app/db/works/works.service.ts
Figments: i'll merge this and change that up in the next PR

PR #167, opened 2020-09-21
Config and Readme updates

pingzing:
NOTE: This PR moves the Render `testing.Dockerfile` and `production.Dockerfile`. I've adjusted the Test service (just for this PR), but after merge, we'll want to permanently change Test, and the Production service too.

A number of things, see inline comments below for details.

- Re-enable Docker Compose (only for dev setup)
- Add separate Dockerfile for dev setup
- Put build-specific dockerfiles in dedicated folder
- Re-enable WASM building as part of build step
- Move client output setup out of Dockerfile and into nx workspace


.render/production.Dockerfile
```diff
@@ -23,9 +23,9 @@ RUN yarn global add @angular/cli@10.0.4 \
 # Download rust and add it to the PATH
 RUN curl https://sh.rustup.rs -sSf |  bash -s -- -y
 RUN echo 'source $HOME/.cargo/env' >> $HOME/.bashrc
+ENV PATH "$PATH:/root/.cargo/bin"
```
pingzing, line 26: This ended up being the secret sauce to get `wasm-pack` working.

.render/production.Dockerfile
```diff
@@ -35,8 +35,6 @@ RUN yarn install
 RUN nx build server
 RUN nx build client --prod
 RUN cp .env ./dist/packages/server/.env
-RUN mkdir ./dist/packages/server/static
-RUN cp -r ./dist/packages/client/* ./dist/packages/server/static/
```
pingzing, line 39: This was moved into the nx workspace.json

build-dev.sh
```diff
@@ -19,7 +18,7 @@ printf "Done!\n"
 
 # Copying client to server and moving into it
 printf "Copying client files to server folder for static file serving...\n"
-mkdir ./dist/packages/server/static
+mkdir -p ./dist/packages/server/static
```
pingzing, line 21: Won't fail if the folder already exists

dev-setup.Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM ubuntu:focal
```
pingzing, line 1: This file is almost the same as the production and testing files, with the exception of not doing `yarn install` or `yarn build`  , because of `node_modules` not persisting if `yarn install` is performed in the Dockerfile, for reasons laid out in more detail here: https://stackoverflow.com/questions/30043872/docker-compose-node-modules-not-present-in-a-volume-after-npm-install-succeeds/32785014#32785014

sample.env
```diff
@@ -1,3 +1,10 @@
 NODE_ENV=development
 DATABASE_URL=mongodb://mongo:27017
 JWT_SECRET=<YOUR SECRET HERE>
+
+IMAGES_HOSTNAME=https://your-image-hostname.com
+
+DIGITALOCEAN_SPACES_ENDPOINT=
+DIGITALOCEAN_SPACES_ACCESS_KEY=
+DIGITALOCEAN_SPACES_SECRET=
+DIGITALOCEAN_SPACES_NAME=
```
pingzing, line 10: Bringing this up to date...

workspace.json
```diff
@@ -15,7 +15,7 @@
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
           "options": {
-            "outputPath": "dist/packages/client",
+            "outputPath": "dist/packages/server/static",
```
pingzing, line 18: With this change, you won't be able to `nx serve client` anymore--you'll have to run the backend. It does mean that we don't have to faff about with copying things around in Dockerfiles though.

workspace.json
```diff
@@ -141,7 +141,8 @@
             "outputPath": "dist/packages/server",
             "main": "packages/server/src/main.ts",
             "tsConfig": "packages/server/tsconfig.app.json",
-            "assets": ["packages/server/src/assets"]
+            "assets": ["packages/server/src/assets"],
+            "webpackConfig": "packages/server/webpack.config.js"
```
pingzing, line 145: ...and a re-enabled webpack config for the `wasm-pack-plugin`.

Figments (approved):
Looks good to me 👍

PR #170, opened 2020-09-29
Add better error messaging when trying to report 'no published sections'

### LINKED ISSUE #112: Report "too few sections" more clearly.
Right now, I don't think we check for it explicitly, so it eventually fails somewhere in the DB side, which returns a generic (confusing!) HTTP 500.

pingzing:
Fixes #112.

Exactly what it says on the tin.

Also, updated the readme, because I realized that line was missing a flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ nx build client
 ```
 On some machines, the frontend won't automatically pick up changes and rebuild. In that case, you can try this:
 ```bash
-nx build client --poll=2000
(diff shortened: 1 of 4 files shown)


pingzing:
Wait, I did a dumb. Do not approve, found a problem.

pingzing:
Okay, undid my dumb.

Figments (approved):

PR #180, opened 2020-10-10
Add h3-h6 to the available header styles, and fix in-section header sizes

### LINKED ISSUE #168: Fix header styles in stories
Two interrelated issues here:
- The editor only gives access to `<h1>` and `<h2>`--it should at least offer `<h3>` and `<h4>` as well, ideally everything up through `<h6>`.
- The only header styles were much smaller than the new ones. The new ones are just the website base styles. While appropriate across the site, they're much too large for stories. They should be scaled down to match the old ones.

pingzing:
Closes #168.

An interesting discovery: if we set a component's contents directly by modifying `innerHtml` (as we do for section bodies and authors' notes), Angular's CSS view encapsulation doesn't apply to the contents. It took me forever to figure out why the changes I was making in the `section-page.less` component weren't applying to headers within a section body. Turns out those rules were being stomped by the global header rules, because I guess they get declared later, and our page-local rules weren't being protected by Angular's encapsulation.

### DIFF
--- a/packages/client/src/app/pages/work-page/section-page/section-page.component.html
+++ b/packages/client/src/app/pages/work-page/section-page/section-page.component.html
@@ -38,7 +38,7 @@
                 </div>
             </div>
             <div class="authors-note" *ngIf="sectionData.authorsNote">
-                <h3>A word from the author...</h3>
+                <h2 class="authors-note-header">A word from the author...</h2>
                 <ng-container *ngIf="!sectionData.usesNewEditor">
                     <quill-view [content]="sectionData.authorsNote" sanitize="true" format="json"></quill-view>
                 </ng-container>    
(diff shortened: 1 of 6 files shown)


packages/client/src/app/pages/work-page/section-page/section-page.component.html
```diff
@@ -38,7 +38,7 @@ <h1>{{ sectionData.title }}</h1>
                 </div>
             </div>
             <div class="authors-note" *ngIf="sectionData.authorsNote">
-                <h3>A word from the author...</h3>
+                <h2 class="authors-note-header">A word from the author...</h2>
```
pingzing, line 41: The `h3` -> `h2` change for the author's note box header is purely semantic--it has the same visual appearance, but I think it makes more sense to consider it an `h2` instead of skipping from the section title's `h1` to an `h3`. It's nicer for screen readers, too, according to MDN.

pingzing:
Also, for comparison:

Current website, using new viewer:
image

Current website using old quill view:
image

With this PR, using new viewer:
image

There are some differences in top-bottom padding. Not sure if we want to keep those or not.

Figments (approved):

PR #186, opened 2020-10-13
Enhancement/resolvers

### LINKED ISSUE #166: Clicking on an item in recent history doesn't navigate to that item until the page is refreshed

Figments:
Closes #166 

Also adds resolvers for the work page, blog page, and portfolio page. Resolvers don't seem to work right on pages that have pagination in them, so further work will be required in the future to bring those up to speed.


pingzing (approved):
looks good :+1:

PR #187, opened 2020-10-13
Adds Title class with methods to set page titles, and sets titles for pages

SigmasonicX:
- Excludes work sections and docs pages
- Excludes Dashboard

Issues:
- Unable to get titles for work sections and new sections working properly
- When refresh page, it first displays the title as the default "Offprint" and then correctly sets the title.


packages/client/src/app/shared/global-constants.ts
```diff
@@ -1,3 +1,26 @@
 export class GlobalConstants {
```
pingzing: I think you can get away with just calling this `Constants`; I'd say the "Global" is implied.

packages/client/src/app/shared/global-methods.ts
```diff
@@ -0,0 +1,16 @@
+import { GlobalConstants } from './global-constants';
+
+export class GlobalMethods {
```
pingzing: Two things:

First:
Personal opinion: I'm not a fan of "Util", "Helper", etc rollup files. Names should be descriptive. In this case, I think something like `titles.ts` for the filename and `Titles` for the class name would be more informative here.

Second:
These functions don't actually need to live in a class--the more idiomatic way to define them would be as naked functions, e.g.
```typescript
export function setDefaultTitle(): void {
    // etc
} 
```

If you want the benefit of things being "inside" something else, you can use a namespace, instead!

```typescript
namespace Titles {
  export function setDefaultTitle(): void {
    // etc
  }
}
```

...and just do `import Titles` elsewhere.

pingzing (changes requested):

packages/client/src/app/shared/title.ts
```diff
@@ -0,0 +1,17 @@
...
+    export function setDefaultTitle(): void {
+        document.title = Constants.OFFPRINT;
+    }
+
+    export function setTwoPartTitle(title: string): void {
+        document.title = title + Constants.SEPARATOR + Constants.OFFPRINT;
+    }
+
+    export function setThreePartTitle(title1: string, title2: string): void {
+        document.title = title1 + Constants.SEPARATOR + title2 +
```
pingzing, line 14: suggestion/idea/thought: Javascript/TypeScript support string interpolation, so you could also write this as
```typescript
document.title = `${title1} ${Constants.SEPARATOR}$ ${title2} ${Constants.SEPARATOR} ${Constants.OFFPRINT}`;
```
...which is a little hard to read, and (personal opinion) I think you're better off not defining a separator constant anyway. Then you could do:
```typescript
document.title = `${title1} - ${title2} - ${Constants.OFFPRINT}` ;
```

PR #188, opened 2020-10-15
feature/client-news

### LINKED ISSUE #107: Add basic News Page functionality
Users should be able to...
- [x] View newsposts on the News page

Contributors, Moderators, and Admins should be able to...

- [x] Create new newsposts
- [x] Edit newsposts, but...
- [x] Only the newspost author can edit their newsposts
- [x]  Preview newsposts

Figments:
Closes #107


packages/client/src/app/pages/news/news.component.html
```diff
@@ -1,12 +1,39 @@
...
-<div style="text-align: center; margin-top: 2rem;">
-    <h3 style="margin-bottom: 0;">Pardon the dust!</h3>
-    <p style="margin-top: 0; margin-bottom: 0; font-style: italic;">Looks like you came here a bit too early.</p>
-    <p style="margin-top: 0; font-style: italic;">Keep an eye on the patch notes to find out when this feature will be available!</p>
-</div>
\ No newline at end of file
+<ng-container *ngIf="posts.totalDocs === 0; else hasDocs">
+    <div style="text-align: center; margin-top: 2rem;">
+        <h3 style="margin-bottom: 0;">Surprisingly, there's nothing here.</h3>
+        <p style="margin-top: 0; margin-bottom: 0; font-style: italic;">Contact an admin if you think something should be here.</p>
```
pingzing: Would these styles be better off as classes? I feel like these get reused in a few places.

packages/client/src/app/services/content/news.service.ts
```diff
@@ -0,0 +1,47 @@
...
+   * Fetches a page of news results.
+   * 
+   * @param pageNum The current page
+   */
+  public getNewsFeed(pageNum: number) {
+    return this.http.get<PaginateResult<NewsContentModel>>(`${this.url}/news-feed/${pageNum}`, {observe: 'response', withCredentials: true})
+      .pipe(map(res => {
+        return res.body;
+      }), catchError(err => {
+        this.snackBar.open(`${err.error.message}`);
```
pingzing, line 28: A thought: I see you're using the material SnackBar now instead of the AlertsService. Should we just wrap the MatSnackBar in the AlertsService?

packages/server/src/app/db/comments/comments.service.ts
```diff
@@ -56,10 +58,31 @@ export class CommentsService {
...
+     * @param commentInfo The comment's info
+     */
+    async createContentComment(user: any, contentId: string, commentInfo: models.CreateComment): Promise<documents.ContentCommentDocument> {
+        const newComment = new this.contentCommentModel({
+            user: user.sub,
+            contentId: contentId,
+            body: commentInfo.body
+        });
+
+        return await newComment.save().then(async doc => {
```
pingzing: don't use a `.then()` callback, they're old and gross. Just do 

```typescript
let doc = await newComment.save();
await this.contentService.addComment(contentId);
return doc;
```
It's semantically equivalent.

packages/server/src/app/db/content/content.service.ts
```diff
@@ -1,10 +1,111 @@
...
+    /**
+     * Finds a bunch of content filtered by kind
+     * 
+     * @param kind The kind of doc
+     * @param pageNum The page number to grab
+     * @param published Optional parameter, checks whether item needs to be published
+     */
+    async fetchManyByKind(kind: string, pageNum: number, published?: boolean):  Promise<PaginateResult<ContentDocument>> {
+        if (published) {
+            if (published === true) {
```
pingzing: I think you can simplify this if-chain to just 
```typescript
if (published === true) { ....
```
...because if `published` is null or `undefined`, that'll evaluate false anyway.

pingzing (changes requested):

pingzing (approved):

Issue #204, opened 2020-10-18
Revamp Portfolio pages


Figments:
The following things are gonna need to be done:

- [x] Change entire look and feel of portfolios
- [x] Migrate My Works, Collections, and My Blogs to portfolio
    - [x] My Works
    - [x] My Blogs
    - [x] Collections
- [x] Increase character limits on bios

...and any other things I figure out we'd need

PR #205, opened 2020-10-19
Enhancement/portfolio revamp

Figments:
Began work on the portfolio revamp outlined in #204 , and migrated the My Works page over. This isn't gonna be the last PR on this.


packages/client/src/app/app-routing.module.ts
```diff
@@ -46,10 +44,12 @@ const routes: Routes = [
     {path: 'portfolio/:id/:username', resolve: {portData: PortfolioResolver}, runGuardsAndResolvers: 'always', component: PortfolioComponent, children: [
       {path: 'blog', component: PortBlogComponent, resolve: {feedData: PortBlogsResolver}, runGuardsAndResolvers: 'always'},
       {path: 'blog/:blogId', resolve: {blogData: BlogPageResolver}, runGuardsAndResolvers: 'always', component: PortBlogPageComponent},
-      {path: 'works', component: PortWorksComponent, resolve: {feedData: PortWorksResolver}, runGuardsAndResolvers: 'always'},
+      {path: 'works', component: WorksComponent, resolve: {feedData: PortWorksResolver}, runGuardsAndResolvers: 'always'},
```
SigmasonicX, line 47: Just noting that the naming convention is inconsistent. Some components and feedDatas both start with "Port", some don't.

Figments: yeah, with this update, a lot of names and components will be normalized so as not to cause as many conflicts

packages/client/src/app/models/site/port-works.ts
```diff
@@ -3,5 +3,5 @@ import { Work } from '@pulp-fiction/models/works';
 
 export interface PortWorks {
     works: PaginateResult<Work>,
-    userId: string;
+    userWorks: PaginateResult<Work>
```
SigmasonicX, line 6: What's the difference between works and userWorks?

Figments: `works` is for the published, content-filtered stuff that anyone other than the author would see
`userWorks` is for all a user's owned works

they both need to come along so that the page displays properly

packages/client/src/app/pages/portfolio/portfolio.component.html
```diff
@@ -1,62 +1,122 @@
...
+            <button class="control right" matTooltip="Report" matTooltipClass="offprint-tooltip"><i-feather name="flag"></i-feather></button>
+        </div>
+        <div class="port-nav">
+            <ul class="navigation-list">
+                <li>
+                    <a [routerLink]="['home']" routerLinkActive="router-link-active">
+                        <span class="link-icon">
+                            <i-feather name="home"></i-feather>
+                        </span>
+                        Home
```
SigmasonicX, line 37: For the future, I recommend using strings from the Constants file instead of writing them out each time.

SigmasonicX (approved):
Left a few comments, but they're more for future changes.

PR #217, opened 2020-10-24
Feature/notifications

pingzing:
Finally ready for review!
There are some major caveats. All the pieces are there, _but_:
 - The only thing that will actually send a notification _right at this moment_ is someone commenting on a blog. That sends a notification to the blog's author. Everything else is pending the content system update being ready.
 - _None of the frontend work is done_. This is pure backend. The controller is ready to talk to the front-end, but that API is totally untested, and may actually be horrible to use.
- I had to hack apart PortfolioController a little to do some testing. I think I restored it, but keep an eye out.


packages/models/src/lib/notifications/notification-kind.ts
```diff
@@ -0,0 +1,9 @@
+export enum NotificationKind {
```
pingzing, line 1: As far as I can figure out, these are all the categories of kinds of notifications. Am I missing anything here?

packages/models/src/lib/notifications/notification.model.ts
```diff
@@ -0,0 +1,27 @@
...
+     * The ID of the thing (Work, Document, Blog, etc) that triggered this notification.
+     */
+    sourceId: string;
+    
+    /**
+     * The ID of the notification source's parent (i.e. if the notification
+     * was triggered by a Section, this will be the parent Work's ID).
+     * Undefined if the source doesn't have a parent.
+     */
+    sourceParentId?: string | undefined;
```
pingzing: The idea behind this `sourceParentId` is that the source of a "new section" notification would be the work. However, that work has a parent (a user) which we probably also want to be able to show in the notification UI.
Ditto new comments (actual parent is a section, or possibly a work, but the parent of _that_ is either a work or a user), etc , etc.

packages/models/src/lib/notifications/unpublished-notification.model.ts
```diff
@@ -0,0 +1,22 @@
+import { NotificationKind } from './notification-kind';
+
+export interface UnpublishedNotification {
+    _id: string;
+
+    /**
+     * The ID of the thing (Work, Document, Blog, etc) that triggered this notification.
+     */
+    sourceId: string;
+
```
pingzing: The only real difference between an `UnpublishedNotification` and a `Notification` is the existence of the a `destinationUserId`--basically, a recipient.

packages/server/src/app/api/notification/notification.controller.ts
```diff
@@ -0,0 +1,14 @@
+import { Controller, Post } from '@nestjs/common';
+
+import { CreateNotification } from '@pulp-fiction/models/notifications';
+import { NotificationsService } from '../../db/notifications/notifications.service';
+
+@Controller('notification')
+export class NotificationController {
+    constructor(private readonly notificationsService: NotificationsService) {}
```
pingzing, line 12: The controller is directly using the DB's `NotificationService`. I remember that most everywhere else we have an extra layer of indirection--an API-level `<Thing>Service` that then hooks into DB-level services if it needs to. Not really sure we need that extra layer here?

packages/server/src/app/db/notifications/notifications.module.ts
```diff
@@ -0,0 +1,38 @@
...
+        name: 'NotificationSubscription',
+        useFactory: () => {
+          const schema = NotificationSubscriptionSchema;
+          return schema;
+        }
+      }
+    ])
+  ],  
+  providers: [NotificationsService, UnpublishedNotificationsService],
+  exports: [NotificationsService],
```
pingzing, line 40: `UnpublishedNotificationService` not being exported is intentional--it's an implementation detail. I may even fold its functionality into the `NotificationService` itself, I'm not sure yet.

packages/server/src/app/db/notifications/notifications.schema.ts
```diff
@@ -0,0 +1,34 @@
...
+    title: string;
+
+    @Prop()
+    body?: string | undefined;
+
+    @Prop({required: true, default: Date.now()})
+    createdAt: Date;
+
+    @Prop({required: true, default: Date.now()})
+    updatedAt: Date
```
pingzing: The `default` automatically gets invoked if this Document has `save()` called on it and that field is `null` or `undefined`, right?

packages/server/src/app/db/notifications/notifications.service.ts
```diff
@@ -0,0 +1,38 @@
...
+        async getAllNotifications(userId: string): Promise<Notification[]> {
+            throw new Error("Not implemented yet");
+        }
+
+        async subscribe(userId: string, sourceId: string): Promise<void> {
+            throw new Error("Not implemented yet");        
+        }
+
+        async unsubscribe(userId: string, sourceId: string): Promise<void> {
+            throw new Error("Not implemented yet");
```
pingzing: The nice thing here is that `subscribe()` and `unsubscribe()` can be super-simple-- we don't _care_ what kind of thing a user is subscribing to. It's the individual notifications' jobs to tell us what `Kind` of thing they are (which we can then use to determine what what the URL is going to look like. 

My thinking is that the client gets a `Notification` object and can use a combination of our known URL schemes, the `Kind` field and the `sourceId` field to construct a URL to the thing that triggered the notification.

...which makes me realize that we'll need a `sourceParentKind` as well as a `sourceParentId`. hm...

packages/server/src/app/db/notifications/unpublished-notifications.service.ts
```diff
@@ -0,0 +1,63 @@
...
+export class UnpublishedNotificationsService {
+    constructor(@InjectModel('UnpublishedNotification') private readonly unpubNotifModel: Model<UnpublishedNotificationDocument>,
+        @InjectModel('NotificationSubscription') private readonly subscriptionModel: Model<NotificationSubscriptionDocument>,
+        @InjectModel('Notification') private readonly notificationModel: Model<NotificationDocument>) {
+
+        // Set up per-tick callback
+        setImmediate(async () => await this.processNotifications(100));
+    }
+
+    async processNotifications(numToProcess: number) {
```
pingzing: This whole service is really just an excuse for this one eternally-looping method. That's one reason I'm thinking of folding it into the `NotificationsService`.

packages/server/src/app/db/notifications/unpublished-notifications.service.ts
```diff
@@ -0,0 +1,63 @@
...
+         performance.measure('processNotifications one iteration', 'processNotifications start', 'processNotifications end');
+         const perfEntry = performance.getEntriesByName('processNotifications one iteration')[0];
+         console.log(`Processed ${toPublish.length} notifications. Afterward, deleted ${deleteResult.deletedCount} unpublished notifications.\nExecution time: ${perfEntry.duration}ms`);
+         performance.clearMeasures('processNotifications one iteration');
+
+        // Queue up another run once this one completes
+        setTimeout(
+            async () => await this.processNotifications(numToProcess),
+            100 // ms to sleep between iterations
+        );
```
pingzing: Using a recursively-called `setTimeout()` here instead of a while loop because 
1) `setTimeout()` called recursively doesn't blow up the call stack, it just schedules a callback for `n` milliseconds in the future
2) The "fire this callback in `n` milliseconds" behavior gives us some nice knobs to twiddle re: speed and performance.

packages/models/src/lib/notifications/mark-read-request.model.ts
```diff
@@ -0,0 +1,3 @@
+export interface MarkReadRequest {
```
pingzing, line 1: This filename doesn't really follow our naming conventions, but just calling it `MarkRead` is kind of vague. Thoughts?

packages/models/src/lib/notifications/publish-status.ts
```diff
@@ -0,0 +1,5 @@
+export enum PublishStatus {
```
pingzing: This has to be a numeric enum, because if we use strings, the enum won't work in a capped collection. They impose a restriction that document sizes can't change, and unless all our enum variants are the exact same string length, changing the enum value will change the document size.

packages/server/src/app/db/comments/comments.service.ts
```diff
@@ -34,6 +37,16 @@ export class CommentsService {
...
         let doc = await newComment.save();
         await this.blogsService.addComment(blogId);
+
+        // Send notification to blog author
+        await this.notificationsService.queueNotification({
+            sourceId: doc._id,
+            sourceKind: NotificationSourceKind.Comment,
+            sourceParentId: blogId,
+            sourceParentKind: NotificationSourceKind.Blog,
+            title: `${user.username} posted a new comment on your blog!` // todo: add in blog title here?
```
pingzing: This is one of the areas where the current data model kind of breaks down. It feels silly to stop, and go make a DB request for both:
1) The commenter's user name
2) The blog's title
...so I really feel like I need to spend some more time thinking about what a client actually needs to construct a useful notification.

packages/server/src/app/db/works/works.service.ts
```diff
@@ -386,12 +402,26 @@ export class WorksService {
...
+    async approveWork(workId: string, authorId: string): Promise<void> {        
+        const workToApprove = await this.workModel.findOne({
+            _id: workId, 
+            //@ts-ignore
+            author: authorId,
+            'audit.isDeleted': false
+        });           
+        workToApprove.audit.published = models.ApprovalStatus.Approved;
+        workToApprove.audit.publishedOn = new Date();
+        await workToApprove.save();
```
pingzing, line 402: This little bit of code could be an example for how we might "fix" the other usages of `update<etc>()`

packages/server/src/app/db/notifications/notification-queue.schema.ts
```diff
@@ -0,0 +1,91 @@
...
+    
+    @Prop()
+    createdAt: Date;
+    
+    @Prop()
+    updatedAt: Date;
+
+    // Why not just use NotificationKind as our discriminator?
+    // Because MongoDB discriminator names must be globally unique.
+    public get kind(): NotificationKind {
```
pingzing: I'm so sorry

packages/client/src/app/components/comments/comments.component.ts
```diff
@@ -157,8 +158,17 @@ export class CommentsComponent implements OnInit {
       return;
     }
 
+    const contentKind: ContentKind = 
+      this.itemKind === ItemKind.Blog ? ContentKind.BlogContent
+      : this.itemKind === ItemKind.BlogContent ? ContentKind.BlogContent
+      : this.itemKind === ItemKind.Work ? ContentKind.WorkContent
+      : this.itemKind === ItemKind.WorkContent ? ContentKind.WorkContent
+      : this.itemKind === ItemKind.NewsContent ? ContentKind.NewsContent
+      : ContentKind.WorkContent;
```
pingzing, line 167: I hacked this in kind of haphazardly, just so I could test comments. Feel free to yell at me to fix this.

packages/models/src/lib/comments/item-kind.enum.ts
```diff
@@ -1,5 +1,7 @@
 export enum ItemKind {
     Blog = 'Blog',
     Work = 'Work',
-    Content = 'Content'
+    BlogContent = 'BlogContent',
+    WorkContent = 'WorkContent',
+    NewsContent = 'NewsContent',
```
pingzing, line 6: Ditto these three.

packages/server/src/app/db/comments/comments.service.ts
```diff
@@ -34,6 +37,19 @@ export class CommentsService {
 
         let doc = await newComment.save();
         await this.blogsService.addComment(blogId);
+
+        const blogTitle = (await this.contentService.fetchOne(blogId, commentInfo.commentParentKind)).title;
+        const notification: CreateCommentNotification = {
+            kind: NotificationKind.CommentNotification,
+            sourceId: blogId,
```
pingzing, line 44: Originally, I had `sourceId` as the ID of the newly-created comment. I soon realized that wouldn't make sense, because there was no way anyone could be subscribed to a comment that _didn't yet exist_.

Fundamentally, `sourceId` should be "ID of the thing people are actually subscribed to".

packages/server/src/app/db/content/blogs/blogs.service.ts
```diff
@@ -30,7 +33,12 @@ export class BlogsService {
             'stats.words': await countPlaintextWords(await stripAllHtml(blogInfo.body))
         });
 
-        return await newBlog.save();
+        const savedBlog = await newBlog.save();
+
+        // Subscribe the author to comments on their new blog
+        await this.notificationsService.subscribe(user.sub, savedBlog._id, NotificationKind.CommentNotification);
```
pingzing, line 39: ...and this is why `sourceId` for comments had to be the ID of the blog. Authors will get notifications for any `CommentKind` notifications with from a source of <their new blog's ID>.

packages/server/src/app/db/notifications/notification-enum-converters.ts
```diff
@@ -0,0 +1,114 @@
+import { NotificationKind } from '@pulp-fiction/models/notifications';
+import { NotificationQueueDocumentKind } from "./notificationQueue/notification-queue-document-kind";
+import { NotificationDocumentKind } from './publishedNotifications/notification-document-kind';
+
+// Contains a variety of helper functions to converted between document-specific
+// discriminators and NotificationKind.
+export namespace NotificationEnumConverters {
```
pingzing, line 7: I hate this entire file.

packages/server/src/app/db/notifications/notifications.module.ts
```diff
@@ -0,0 +1,63 @@
...
+        useFactory: () => {
+          const schema = NotificationSubscriptionSchema;
+          return schema;
+        }
+      }
+    ])
+  ],  
+  providers: [NotificationsService,  
+    ...NotificationQueueSubSchemaProviders,  
+    ...NotificationSubSchemaProviders
```
pingzing, line 38: These two lines involve some intense voodoo.

packages/server/src/app/db/notifications/notifications.service.ts
```diff
@@ -0,0 +1,415 @@
...
+            { publishStatus: PublishStatus.InProgress, updatedAt: { $lte: nowMinus10s } }
+        );
+        await Promise.all(staleNotifications.map(x => {
+            x.publishStatus = PublishStatus.NotStarted;
+            return x.save();
+        }));
+    }
+
+    private getSpecificNotificationModel(subscription: NotificationSubscriptionDocument, toPublish: NotificationQueueDocument): NotificationDocument {
+        const commonProperties: NotificationBase = {
```
pingzing, line 313: This method and those below are the last bit of really gnarly boilerplate. I tried a few different ways to cut it down to something a little more OOPy, or less horrible, to no avail. This was the least-worst I could get it.

packages/server/src/app/db/works/works.schema.ts
```diff
@@ -1,77 +1,79 @@
...
+
+WorksSchema.index({title: 'text'});
+
+WorksSchema.plugin(MongooseAutopopulate);
+WorksSchema.plugin(MongoosePaginate);
+
+WorksSchema.pre<documents.WorkDocument>('save', async function (next: HookNextFunction) {
+    if (!this._id) {
+        this.set('_id', generate());
+    }
```
pingzing, line 62: These lines are the only ones that actually changed. Not sure why git is calling out the whole file.

Figments (approved):

PR #221, opened 2020-11-15
Mobile optimization - Resize navbar with different screen widths

SigmasonicX:
Change the navbar font size and spacing according to the screen width, which should make it look better on mobile devices.

Note that variables can't be used in media fields, but I choose to include them in styles.less for reference.

### DIFF
--- a/packages/client/src/app/app.component.less
+++ b/packages/client/src/app/app.component.less
@@ -119,7 +119,7 @@ nav {
   max-width: 80%;
   margin: 0 auto 0 auto;
 
-  @media (max-width: 1150px) {
+  @media (max-width: 1150px) { //--xlarge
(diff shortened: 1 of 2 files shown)


pingzing (approved):
Looks good. For the `xsmall` and `small` sizes, you may even consider getting rid of the text altogether.

SigmasonicX:
> 
> 
> Looks good. For the `xsmall` and `small` sizes, you may even consider getting rid of the text altogether.

You mean the comments? I thought as a work-around for variables in media(), it made sense to comment with the "variable" name, so when we come up with a better solution, we can search for those variable names.

pingzing:
Oh, no, I mean the text next to the icons in the navbar.

PR #225, opened 2020-11-26
Enhancement/content api prose publishing

Figments:
Adds in the prose page with section navigation

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "@types/sanitize-html": "^1.23.3",
     "@types/shortid": "^0.0.29",
     "@types/uuid": "^8.3.0",
+    "@types/voca": "^1.4.0",
     "angular-feather": "^6.0.10",
     "argon2": "^0.27.0",
     "aws-sdk": "^2.738.0",
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "sanitize-html": "^1.27.3",
     "shortid": "^2.2.15",
     "validator": "^13.1.1",
+    "voca": "^1.4.0",
     "zone.js": "^0.10.2"
   },
   "devDependencies": {
--- a/packages/client/src/app/components/rating-icon/rating-icon.component.ts
+++ b/packages/client/src/app/components/rating-icon/rating-icon.component.ts
@@ -9,6 +9,7 @@ import { ContentRating } from '@pulp-fiction/models/content';
 })
 export class RatingIconComponent implements OnInit {
     @Input() rating: ContentRating;
+    @Input() size: string;
 
     constructor() {}
 
--- a/packages/client/src/app/components/work-card/work-card.component.ts
+++ b/packages/client/src/app/components/work-card/work-card.component.ts
@@ -1,0 +1,21 @@
+import { Component, Input, OnInit } from '@angular/core';
+
+import { calculateApprovalRating } from '../../util/functions';
+
+@Component({
+    selector: 'work-card',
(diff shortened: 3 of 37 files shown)


packages/client/src/app/components/rating-icon/rating-icon.component.ts
```diff
@@ -9,6 +9,7 @@ import { ContentRating } from '@pulp-fiction/models/content';
 })
 export class RatingIconComponent implements OnInit {
     @Input() rating: ContentRating;
+    @Input() size: string;
```
pingzing, line 12: Super minor nitpick: this could be an enum like RatingIconSize or something (with appropriate string representations)

packages/client/src/app/components/work-card/work-card.component.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { calculateApprovalRating } from '../../util/functions';
+
+@Component({
+    selector: 'work-card',
+    templateUrl: './work-card.component.html',
+    styleUrls: ['./work-card.component.less']
+})
+export class WorkCardComponent implements OnInit {
+    @Input() content: any;
```
pingzing, line 14: huh, can we not make this a more specific type? coulda sworn we had a base Content interface

packages/client/src/app/pages/content-views/section-view/section-view.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    constructor(private contentService: ContentService, private route: ActivatedRoute, private router: Router) {
+        this.fetchData();
+    }
+
+    ngOnInit(): void {}
+
+    private fetchData() {
+        this.loading = true;
+        this.route.paramMap.subscribe(params => {
```
pingzing, line 30: Can you use the `paramMap.snapshot` here? Or do you actually need a persistent subscription?

packages/client/src/app/pages/content-views/section-view/section-view.component.ts
```diff
@@ -0,0 +1,56 @@
+import { Component, OnInit } from '@angular/core';
+import { ContentService } from '../../../services/content';
+import { ContentModel, SectionInfo } from '@pulp-fiction/models/content';
+import { Section } from '@pulp-fiction/models/sections';
+import { ActivatedRoute, Router } from '@angular/router';
+import * as v from 'voca';
```
pingzing: Importing the entire `voca` library just for their `slugify()` function strikes me as _super_ wasteful. I am making some assumptions about size though--what's the difference in compiled client size?

Voca's MIT-licensed--we could probably just lift their implementation without bringing the whole library along for the ride.

packages/client/src/app/pages/portfolio/works/works.component.ts
```diff
@@ -46,8 +45,8 @@ export class WorksComponent implements OnInit {
 
         this.route.data.subscribe(data => {
```
pingzing, line 46: While you're changing this file anyway, can you get away with changing this `route.data.subscribe` to `route.data.snapshot`?

package.json
```diff
@@ -95,6 +95,7 @@
     "sanitize-html": "^1.27.3",
     "shortid": "^2.2.15",
     "validator": "^13.1.1",
+    "voca": "^1.4.0",
```
pingzing, line 99: I have Opinions about including this--see below

pingzing (changes requested):

pingzing (approved):

PR #227, opened 2020-11-28
Fix/double viewcounts

pingzing:
The issue was caused by the Comments component paging triggering the resolver again. Now resolvers should only fire if route parameters change.


Figments (approved):

SigmasonicX (approved):
LGTM

PR #233, opened 2020-12-13
Use locale dates

SigmasonicX:
https://github.com/OffprintStudios/pulp-fiction/issues/109
- Set slogan to use locale date
- Add localedate pipe, which replaces all uses of the date pipe in HTML


pingzing (changes requested):
For your new pipe to work, it also needs to be declared in `app.module.ts`'s `@NGModule()` `declarations` array.

pingzing (approved):
Looks good. I'd suggest adding some error handling to the pipe (what happens if a dev screws up and passes some arbitrary string in?), but that's a low-enough impact issue that I'm fine approving this 👍

SigmasonicX:
The default case can handle mistakes in the string passed in, and the Date class can handle getting bad strings; toLocaleDateString() will return "Invalid Date" as a string, and won't crash.

PR #235, opened 2020-12-19
Update Browse page to use Content API

pingzing:
Also updated the `separateEntites` pipe. It's just used for separating Genres now, so it's been renamed. We also needed a way to turn a `string[]` of fandoms into a comma-separated string, so I introduced a general `JoinStringsPipe` that does just that for any array of strings.

EDIT: Part 2 - Done! Had to do some backend work to make the `get-all-published` understand how to handle a client asking for an array for `ContentKind`s, but it was a small change, and makes the client logic significantly simpler. Also had to do a bit of cleanup around category pipes and such.

Request to @Figments: can you double check the component that Browse page links take users to? I'm pretty sure it's the right one, but I'm not 100% what your plan is there.

Status: Ready for review!

### DIFF
--- a/packages/client/src/app/pages/browse/browse.component.html
+++ b/packages/client/src/app/pages/browse/browse.component.html
@@ -20,14 +20,14 @@
             <div class="work-container">
                 <!--Header-->
                 <div class="work-header">
-                    <rating-icon [rating]="work.meta.rating" [size]="small"/>                          
+                    <rating-icon [rating]="work.meta.rating" [size]="small"></rating-icon>                          
                     <h1><a [routerLink]="['/work', work._id, work.title | slugify]">{{ work.title }}</a></h1>                            
                     <h2>by <a [routerLink]="['/portfolio', work.author._id, work.author.username | slugify]">{{ work.author.username }}</a></h2>
                     <div class="approval" title="Approval Rating">
                         <i-feather name="heart"></i-feather>{{ calcApprovalRating(work.stats.likes, work.stats.dislikes) }}%
                     </div>
                     <div class="category-genre">
-                        {{ work.meta.category | fixCategories }}<span>//</span>{{ work.meta.genres | separateEntities }}<ng-container *ngIf="work.meta.fandoms.length > 0"><span>//</span>{{ work.meta.fandoms | separateEntities }}</ng-container>
+                        {{ work.meta.category | fixCategories }}<span>//</span>{{ work.meta.genres | separateGenres }}<ng-container *ngIf="work.meta.fandoms.length > 0"><span>//</span>{{ work.meta.fandoms | joinStrings }}</ng-container>
                     </div>                            
                 </div>
 
@@ -37,7 +37,7 @@
                         <img class="cover-art bordered-cover-art" [src]="work.meta.coverArt">
                     </div>
                     <div>
-                        {{ work.shortDesc }}
+                        {{ work.desc }}
                     </div>
                 </div>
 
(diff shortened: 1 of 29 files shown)


packages/client/src/app/pages/browse/browse.component.html
```diff
@@ -20,19 +20,17 @@ <h2>Latest Works</h2>
...
-                        <div matTooltip="Teen" matTooltipClass="offprint-tooltip" class="rating teen small" *ngSwitchCase="'Teen'">T</div>
-                        <div matTooltip="Mature" matTooltipClass="offprint-tooltip" class="rating mature small" *ngSwitchCase="'Mature'">M</div>
-                        <div matTooltip="Explicit" matTooltipClass="offprint-tooltip" class="rating explicit small" *ngSwitchCase="'Explicit'">X</div>                                
-                    </ng-container>                            
-                    <h1><a [routerLink]="['/work', work._id, work.title | slugify]">{{ work.title }}</a></h1>                            
+                    <rating-icon [rating]="work.meta.rating" [size]="'small'"></rating-icon>                          
+                    <h1 [ngSwitch]="work.kind">
+                        <a *ngSwitchCase="'ProseContent'" [routerLink]="['/prose', work._id, work.title | slugify]">{{ work.title }}</a>
+                        <a *ngSwitchCase="'PoetryContent'" [routerLink]="['/poetry', work._id, work.title | slugify]">{{ work.title }}</a>
```
pingzing, line 26: This is a bit clumsy, but seemed like the best way to handle both prose and poetry.

packages/client/src/app/pipes/fix-categories.pipe.ts
```diff
@@ -1,9 +1,14 @@
 import { Pipe, PipeTransform } from '@angular/core';
+import { WorkKind } from '@pulp-fiction/models/content';
 import { Categories } from '@pulp-fiction/models/works';
 
-@Pipe({name: 'fixCategories'})
+@Pipe({ name: 'fixCategories' })
 export class FixCategoriesPipe implements PipeTransform {
-    transform(value: Categories) {
-        return Categories[value];
+    transform(value: Categories | WorkKind) {
```
pingzing, line 7: Now it works for both old and new categories! whee.

packages/client/src/app/resolvers/work-page.resolver.ts
```diff
@@ -8,6 +8,9 @@ import { WorkPageData } from '../models/site';
 import { AuthService } from '../services/auth';
 import { HistoryService, LocalSectionsService, WorksService } from '../services/content';
 
+
+// This resolver is obsolete. 
+// TODO: Delete.
 @Injectable()
 export class WorkPageResolver implements Resolve<WorkPageData> {
```
pingzing, line 15: I'm a little afraid to remove this outright, so just a TODO for now. We can go through and do a removal pass once the dust from the API change clears.

packages/client/src/app/services/content/content.service.ts
```diff
@@ -26,16 +27,17 @@ export class ContentService {
       }));
   }
 
-  public fetchAllPublished(pageNum: number, kind: ContentKind, userId?: string) {
+  public fetchAllPublished(pageNum: number, kind: ContentKind[], userId?: string): Observable<PaginateResult<ContentModel>> {
     let route = `${this.url}`;
 
+    // If we just include the kind array as-is, it'll be serialized as "&kind=Kind1,Kind2" which the backend will interpret as
+    // the string 'Kind1,Kind2' which is not what we want. So, we manually split it out into a query string
+    const kindFragment = kind.map(k => `&kind=${k}`).join('');
```
pingzing, line 35: I did not expect NestJS to handle this as gracefully as it did. +1 for convenience here

packages/client/src/styles.less
```diff
@@ -134,7 +134,7 @@ div.rating {
         height: 24px;
         width: 24px;
     }
-    &.large {
+    &.medium {
```
pingzing, line 137: had duplicate larges here. this may result in `large`s elsewhere being too big now, because I think they may have been short-circuting on what is now `medium`.

packages/server/src/app/api/content/content.controller.ts
```diff
@@ -38,7 +38,7 @@ export class ContentController {
     }
 
     @Get('fetch-all-published')
-    async fetchAllPublished(@Request() req: any, @Query('pageNum') pageNum: number, @Query('userId') userId: string, @Query('kind') kind: ContentKind) {
+    async fetchAllPublished(@Request() req: any, @Query('pageNum') pageNum: number, @Query('userId') userId: string, @Query('kind') kind: ContentKind[]) {
```
pingzing, line 41: Array of `ContentKind`s now. Will make getting different combinations dead easy.

Figments:
at first glance, I'm not seeing anything related to segregating content based on content rating. maybe I'm just missing something, though.

Figments (approved):

PR #256, opened 2020-12-28
Fix/prose and poetry page display

Figments:
Fixes the following issues with poetry and prose content views:
- Dislike counter using the `likes` stat tracker
- Author byline is now on the same line as the title, just as before

Also bumps the site version to 0.5.2


SigmasonicX (approved):
LGTM

pingzing (approved):

PR #263, opened 2020-12-30
Enhancement/re enable authors notes

Figments:
brings back authors notes


SigmasonicX (approved):
More duplicated code than I'd like, but it's ultimately not too many places.

PR #270, opened 2020-12-30
Fix/poetry bugs

Figments:
Fixes the following issues:
- work cards not displaying the correct content kind
- work cards now have correct linking behavior (i.e. you can now right click and open a fic in a new tab)
- work status now added to prose and poetry pages
- poetry pages now have correct history behavior
- portfolio works feed now combines prose and poetry into one list


packages/client/src/app/pages/content-views/poetry-page/poetry-page.component.ts
```diff
@@ -31,6 +32,7 @@ export class PoetryPageComponent implements OnInit {
     ngOnInit(): void {
         this.route.data.subscribe(data => {
             const pageData = data.poetryData as ContentPage;
+            console.log(pageData);
```
SigmasonicX: Should this still be here?

SigmasonicX (approved):
Aside from the lingering console log, LGTM

PR #272, opened 2020-12-30
Fix/section editing

### LINKED ISSUE #243: After add section, need to refresh page in order to update the work's word count
Currently, after adding a section to a work, the work's word count isn't updated until after you refresh the page. This can cause confusion when you try to submit the work.

### LINKED ISSUE #260: After you edit a section, the preview doesn't reflect this change until you refresh
Currently, when you edit a section and click "Save", the changes you made aren't reflected in the preview until you manually refresh. This makes for a bad user experience.

Figments:
Fixes the following issues:
- Editing a section now repopulates the section item with the new edits
- Saving a section will now report whether the action was successful or not
- Page will reload whenever a section is published so that the work can be refetched (temporary solution)
- Prose and poetry editing pages now have their Status displayed

Also updates all our base packages to their latest versions (nothing broke on my local machine when I did this, so it looks like we're all good unless the test environment fails).

Closes #260 and closes #243


SigmasonicX (approved):
LGTM

PR #285, opened 2021-01-03
Enhancement/loading bar

Figments:
adds a loading bar that triggers on all HTTP requests

### DIFF
--- a/packages/client/src/app/app.component.html
+++ b/packages/client/src/app/app.component.html
@@ -1,4 +1,3 @@
-
 <mat-sidenav-container>
   <mat-sidenav [fixedInViewport]="true" [(opened)]="sidenavOpened" #sidenav>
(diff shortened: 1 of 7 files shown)


packages/client/src/app/app.component.html
```diff
@@ -31,6 +30,8 @@ <h2>{{ rotatingSlogan }}</h2>
         </div>
       </header>
     </div>
+
+    <mat-progress-bar [mode]="'determinate'" [value]="loader.value$ | async"></mat-progress-bar>
```
pingzing, line 34: Huh, what's the `value$` syntax? Haven't seen that before.

And I'm guessing the `async` pipe comes along for the ride with the `LoadingBarHttpClientModule`?

pingzing (approved):

PR #300, opened 2021-01-09
Update search page to work with content API

pingzing:
A few bits of weirdness here. See inline comments.

Note: I updated the `find-<thing>` components, but didn't actually add them to the search page--I figured you had a plan here and didn't want to step on toes.

### DIFF
--- a/packages/client/src/app/pages/search/search.component.html
+++ b/packages/client/src/app/pages/search/search.component.html
@@ -1,5 +1,142 @@
-<div class="empty">
-    <h3>Pardon the dust!</h3>
-    <p>This feature has been temporarily disabled.</p>
-    <p>Keep an eye on the patch notes to see when this is coming back!</p>
+<div class="search-nav">
+    <div class="search-bar">
+        <form [formGroup]="searchForm" (ngSubmit)="submitSearch()">
+            <input formControlName="query" type="text" class="u-full-width" placeholder="Search ...">
+        </form>
+    </div>
 </div>
+
+<ng-container *ngIf="route.children.length === 0">
+    <ng-container *ngIf="searchResults === undefined; else resultsAreIn">
+        <div style="text-align: center;">
+            <h3 style="margin-bottom: 0;">You haven't searched for anything yet...</h3>
+            <p style="margin-top: 0; margin-bottom: 0;">Type your query in the search bar and hit enter.</p>
+            <p style="margin-top: 0;">The first six works/blogs/users will show up here.</p>
+        </div>
+    </ng-container>
+    <ng-template #resultsAreIn>
+        <div class="search-results">
+
+            <!--Works-->
+            <div class="results">
+                <div class="results-header">
+                    <i-feather name="feather"></i-feather>
+                    <h3>Top Works</h3>
+                    <span class="see-more" *ngIf="searchResults.works.length !== 0">
+                        <a [routerLink]="['/search/works']" [queryParams]="{ query: searchField.query.value }" *ngIf="searchResults.works.length > 5">See {{ searchResults.works.length - 6 }} more</a>
+                    </span>
+                </div>
+                <div class="results-body">
+                    <ng-container *ngIf="searchResults.works.length === 0; else hasWorks">
+                        <div style="text-align: center; margin-top: 1.25rem;">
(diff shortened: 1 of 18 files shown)


packages/client/src/app/pages/search/search.component.html
```diff
@@ -1,5 +1,142 @@
...
\ No newline at end of file
+<div class="search-nav">
+    <div class="search-bar">
+        <form [formGroup]="searchForm" (ngSubmit)="submitSearch()">
+            <input formControlName="query" type="text" class="u-full-width" placeholder="Search ...">
+        </form>
+    </div>
+</div>
+
+<ng-container *ngIf="route.children.length === 0">
```
pingzing, line 9: Note about this page: all three search result types are very visually different. I did a little bit of cleanup here, but not much.

packages/client/src/app/pages/search/search.component.less
```diff
@@ -19,9 +19,10 @@ div.search-nav {
             color: whitesmoke;
             box-shadow: none;
             font-size: 32px;
+            height: auto;
             border: 1px solid whitesmoke;
             margin-bottom: 0;
-            padding: 5px 20px 5px 20px;
+            padding: 10px 20px 10px 20px;
```
pingzing, line 25: Giving the search text a bit of room to breathe...

packages/client/src/app/pages/search/search.component.less
```diff
@@ -217,6 +219,7 @@ div.search-results {
 
                     div.profile {
                         width: 100%;
+                        min-width: 0;
```
pingzing, line 222: Required for the username truncation to work

packages/client/src/app/pages/search/search.component.less
```diff
@@ -226,6 +229,9 @@ div.search-results {
                                 font-weight: 500;
                                 font-size: 38px;
                                 margin-bottom: 0;
+                                white-space: nowrap;
+                                text-overflow: ellipsis;
+                                overflow: hidden;
```
pingzing, line 234: So stupid-long usernames don't overflow

packages/models/src/lib/util/initial-results.model.ts
```diff
@@ -0,0 +1,8 @@
+import { ContentModel } from '../content';
+import { User } from '../users';
+
+export interface InitialResults {
+    users: User[],
+    blogs: ContentModel[],
+    works: ContentModel[],
```
pingzing, line 7: This is one bit of weirdness--the initial implementation for searching just hits the content service, so it just gets back ContentModels. I could've implemented searches down in the individual services, but it doesn't look like we need that yet. (if we do in the future, we can always implement it there too)

packages/models/src/lib/util/paginate-result.model.ts
```diff
@@ -2,13 +2,12 @@ export interface PaginateResult<T> {
...
-    readonly offset?: number;
-    readonly prevPage: number;
-    readonly nextPage: number;
+    readonly nextPage?: number;
+    readonly prevPage?: number;
     readonly pagingCounter: number;
-    readonly meta?: {};
+    readonly hasPrevPage: boolean;
+    readonly hasNextPage: boolean;
+    readonly meta?: any;
```
pingzing, line 12: Some extremely minor cleanup to make this match this interface match the mongoose-side type so they're mutually-compatible. Now we don't have to rely on `any`-based duck-typing.

packages/server/src/app/db/users/users.service.ts
```diff
@@ -307,28 +307,18 @@ export class UsersService {
...
-     * Finds the first six matches given the provided search parameters.
-     * For use with the initial page of search results.
-     * 
-     * @param query The relevant search parameters
-     */
-    async findInitialRelatedUsers(query: string): Promise<documents.UserDocument[]>{
-        return await this.userModel.find({$text: {$search: query}})
-            .select('-password -agreedToPolicies -audit.sessions -audit.termsAgree -audit.emailConfirmed -audit.isDeleted')
-            .limit(6);
-    }
```
pingzing, line 320: Folded this into `findRelatedUsers`.

Figments (approved):

PR #301, opened 2021-01-10
Enhancement/navigation refresh

Figments:
Addresses the following:
- Changes style of navigation bar to be far more aesthetically appealing
- Makes navigation bar horizontally scrollable in mobile view
- Removes "News" page link
- Changes "Groups" page link name to "Social"


SigmasonicX (approved):
LGTM

PR #314, opened 2021-01-13
removed buggy style preventing scroll to top behavior from working

### LINKED ISSUE #311: When switching sections, go to the top of the section
User request
Currently, when you switch sections, if you're at the bottom of the page, you remain at the bottom of the page.

Figments:
Closes #311 

Turns out, setting the `html, body` height to 100% actually makes `scrollPositionRestoration` bug out and not scroll to top on route change like it's supposed to.

### DIFF
--- a/packages/client/src/styles.less
+++ b/packages/client/src/styles.less
@@ -39,7 +39,6 @@
 
 html,body {
     width: 100%;
-    height: 100%;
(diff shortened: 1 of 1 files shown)


SigmasonicX (approved):
Worked for me

PR #327, opened 2021-01-18
Enhancement/ngxs transition

Figments:
Begins the transition to using NGXS Store for state management. This migration only deals with authentication and user settings, with further migrations expected to occur over the next series of updates.


package.json
```diff
@@ -58,6 +58,11 @@
     "@ngu/carousel": "^3.0.2",
     "@ngx-loading-bar/core": "^5.1.1",
     "@ngx-loading-bar/http-client": "^5.1.1",
+    "@ngxs/devtools-plugin": "^3.7.1",
```
pingzing, line 61: This should probably be a devDependency, yeah?

packages/client/src/app/app.component.ts
```diff
@@ -21,42 +21,49 @@ import { NotificationBase } from '@pulp-fiction/models/notifications';
...
-
-    // Sets the current site theme based on user preference
+    
     if (this.currentUser) {
-        this.changeTheme(PredefinedThemes[this.currentUser.profile.themePref]);     
+        // Sets the current site theme based on user preference
+        this.changeTheme(PredefinedThemes[this.currentUser.profile.themePref]);   
+
+        // Starts fetching notifications updates from the server
+        interval(300000).pipe(flatMap(() => this.notif.getUnreadNotifications())).subscribe(data => {
```
pingzing: Probably better practice to stick this in a `FIVE_MINUTES` const or something--magic numbers bad.

packages/client/src/app/components/site-sidenav/notifications/notif-item/notif-item.component.html
```diff
@@ -0,0 +1,22 @@
+<div class="notif-item">
+    <a class="notification-link" (click)="sendTo()">
```
pingzing, line 2: Can this be a real link? If it's a "let's do it later", make an issue for it

packages/client/src/app/components/site-sidenav/notifications/notif-item/notif-item.component.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { ContentKind } from '@pulp-fiction/models/content';
+
+@Component({
+    selector: 'notif-item',
+    templateUrl: './notif-item.component.html',
+    styleUrls: ['./notif-item.component.less']
+})
+export class NotifItemComponent implements OnInit {
+    @Input() notif: any; // this is `any` because different notification documents have different fields
```
pingzing, line 14: This feels like a sign that I need to make the interfaces inherit from the NotificationBase interface. And I think I can even do that...

pingzing (changes requested):
Requesting changes from myself re: notifications sharing a common ancestor interface

packages/client/src/app/components/site-sidenav/notifications/notif-item/notif-item.component.ts
pingzing: EDIT: Done. Recommend this be changed from `any` to `NotificationBase`. If you need further specificity, you've already got a `kind` check further down

packages/client/src/app/components/site-sidenav/notifications/notification-select.model.ts
```diff
@@ -0,0 +1,5 @@
+import { NotificationBase } from '@pulp-fiction/models/notifications';
+
+export interface NotificationSelect extends NotificationBase {
```
pingzing, line 3: Nitpick: This is a data model that adds some view state. Do we want to follow a convention where these kinds of things are all called "viewmodels", the way we have elsewhere? or nah?

packages/client/src/app/components/site-sidenav/notifications/notifications.component.ts
```diff
@@ -39,10 +43,31 @@ export class NotificationsComponent implements OnInit {
...
+
+  markAsRead(): void {
+    const request: MarkReadRequest = {
+      ids: this.selectedNotifs
+    };
+
+    this.notif.markAsRead(request).subscribe(() => {
+      this.selectedNotifs = [];
+      this.fetchData();
+    }, () => {
```
pingzing: Can this be `err => { ...` just so it's a little more immediately clear it's the error block, even if we don't do anything with `err`?

packages/client/src/app/components/site-sidenav/site-sidenav.component.ts
```diff
@@ -84,13 +91,14 @@ export class SiteSidenavComponent implements OnInit {
       password: this.loginFields.password.value, 
       rememberMe: this.loginFields.rememberMe.value
     };
-    this.authService.login(credentials).pipe(first()).subscribe(() => {
+
+    this.store.dispatch([new Auth.Login(credentials), new Navigate(['/home'])]).subscribe(() => {
       this.loadingLogin = false;
-      this.router.navigate(['/home']);
+      location.reload();
```
pingzing, line 97: Shouldn't the `Navigate(['/home'])` take care of this? Or am I misunderstanding the `dispatch()` syntax?

packages/client/src/app/models/site/theme.ts
```diff
@@ -41,125 +41,162 @@ export interface Theme {
...
-    accentLight: '#E58E8E',
-    background: '#272727',
-    textColor: '#F5F5F5', // aka WhiteSmoke
-    borders: '#FFFFFF', // aka White
-    controlsBackground: '#3A3A3A',
-    codeBackground: '#3A3A3A',
+/*
+export const PredefinedThemes = Record<string; Theme> = {
+    'crimson' = Crimson;
+    'dark-crimson' = DarkCrimson;
```
pingzing: I think this is safe to delete. no need to keep it around

packages/client/src/app/shared/auth/services/auth.guard.ts
```diff
@@ -0,0 +1,81 @@
...
+})
+export class AuthGuard implements CanActivate, CanActivateChild {
+    helper = new JwtHelperService();
+
+    constructor(private store: Store, private snackBar: MatSnackBar) {}
+
+    /**
+     * Verifies that a user can access a protected route. If their JWT is expired, it makes a request to the backend to
+     * verify that an active session is still present. If one is, the route request is approved. If one isn't, then the
+     * request is denied.
```
pingzing, line 23: Is the part about requesting the backend here still true?

packages/client/src/app/components/site-sidenav/notifications/notif-item/notif-item.component.html
Figments: that would introduce more `ng-container`s and `ngSwitch`es and the reason notifications got moved to their own subcomponent was to combat an issue where too many was preventing certain elements from even *appearing* on the page

packages/client/src/app/app.component.ts
Figments: true

packages/client/src/app/shared/auth/services/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+})
+export class AuthService {
+    private url: string = `/api/auth`;
+
+    constructor(private http: HttpClient) {}
+
+    /**
+     * Sends a new user's info to the backend. If the response is successful,
+     * the `user` is returned and their info is set in local storage. Otherwise,
+     * sends the backend error to all subscribed observables.
```
pingzing, line 24: Is the bit about setting the info in local storage still true?

packages/client/src/app/shared/auth/services/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+                return user.body;
+            }), catchError(err => {
+                return throwError(err);
+            }));
+    }
+
+      /**
+   * Sends a returning user's credentials to the backend to be verified.
+   * If the response is successful, the `user` is returned and their info
+   * is set in local storage. Otherwise, sends the backend error to all
```
pingzing: ditto local storage again

packages/client/src/app/shared/auth/services/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+            .pipe(map(user => {
+                return user.body;
+            }), catchError(err => {
+                return throwError(err);
+            }));
+    }
+
+    /**
+     * Logs the user out, sets the user object to null, removes their info from localStorage, and
+     * navigates to home.
```
pingzing: I don't think this function does most of this anymore--it's just an HTTP request now

packages/client/src/app/shared/auth/services/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public logout(): Observable<void> {
+        return this.http.get<void>(`${this.url}/logout`, { withCredentials: true });
+    }
+
+    /**
+     * Refreshes the current user token with new User info.
+     * If refresh fails,
```
pingzing: > If refresh fails,

...then what?

packages/client/src/app/shared/global/global.state.ts
```diff
@@ -0,0 +1,31 @@
...
+import { ContentFilter } from '@pulp-fiction/models/content';
+
+@State<GlobalStateModel>({
+    name: 'global',
+    defaults: {
+        filter: ContentFilter.Default
+    }
+})
+@Injectable()
+export class GlobalState {
```
pingzing, line 18: Would a better name for this stuff be something like "UserPreferences", or maybe even "VisitorPreferences" to distinguish that this is for _all_ visitors to the site, instead of just logged-in ones?

packages/client/src/app/shared/global/services/global.service.ts
```diff
@@ -0,0 +1,38 @@
+import { Injectable } from '@angular/core';
+import { CookieService, CookieOptions } from 'ngx-cookie';
+
+import { ContentFilter } from '@pulp-fiction/models/content';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class GlobalService {
```
pingzing, line 9: see above re: naming

packages/models/src/lib/users/theme.ts
```diff
@@ -0,0 +1,165 @@
...
+export const Crimson: Theme = {
+    accent: '#DD4C4F',
+    accentHover: '#933235',
+    accentLight: '#EA7C7C',
+    background: '#FBFBFB',
+    textColor: '#000000', // aka black
+    borders: '#808080', // aka grey
+    controlsBackground: '#D3D3D3', // aka lightgrey
+    codeBackground: '#F0F0F0',
+}
```
pingzing, line 53: Wait, aren't these also defined elsewhere?

packages/models/src/lib/users/theme.ts
```diff
@@ -0,0 +1,165 @@
...
+    'crimson': Crimson,
+    'dark-crimson': DarkCrimson,
+    'aqua': Aqua,
+    'dark-aqua': DarkAqua,
+    'royal': Royal,
+    'dark-royal': DarkRoyal,
+    'steel': Steel,
+    'midnight-field': MidnightField,
+    'autumn': Autumn,
+    'dusk-autumn': DuskAutumn
```
pingzing, line 164: yeah, this whole file is a dupe, I think

packages/server/src/app/api/auth/auth.service.ts
```diff
@@ -5,6 +5,7 @@ import { verify, argon2id } from 'argon2';
 import { User, FrontendUser, ChangePassword, ChangeProfile, ChangeEmail, ChangeUsername, UpdateTagline } from '@pulp-fiction/models/users';
 import { UsersService } from '../../db/users/users.service';
 import { JwtPayload } from '@pulp-fiction/models/auth';
+import { String } from 'aws-sdk/clients/appstream';
```
pingzing: wait, for real?

pingzing (changes requested):
A few things that should be fixed! Mostly small stuff, though. And I have to admit--I like `ngxs`--it does make state management a lot more cleanly separated.

One thing that I noticed that we could do to clean things up a lot: now that our state management and network calls are divorced from each other, we should move all our little services that do nothing but make network calls into a single `networkService`. That'd make it easier to have little shared helper functions, and we wouldn't have a billion tiny services everywhere that I always lose track of =P

packages/client/src/app/components/site-sidenav/notifications/notif-item/notif-item.component.html
pingzing: fair, I guess. but I _guarantee_ this is gonna piss users off

packages/client/src/app/components/site-sidenav/notifications/notifications.component.ts
Figments: i could tag it `_err` so it does pop up with "unused variable" warnings

packages/client/src/app/components/site-sidenav/site-sidenav.component.ts
Figments: see conversation in discord

packages/client/src/app/shared/auth/services/auth.guard.ts
Figments: yes

packages/client/src/app/shared/auth/services/auth.service.ts
Figments: yes--`ngxs/storage-plugin` automatically takes care of setting things in `localStorage`

packages/client/src/app/shared/auth/services/auth.service.ts
Figments: see last comment

packages/client/src/app/shared/auth/services/auth.service.ts
Figments: yeah, will update the docs

packages/client/src/app/shared/auth/services/auth.service.ts
Figments: lmao

packages/models/src/lib/users/theme.ts
Figments: yeah i just copied over the theme stuff from the client

packages/server/src/app/api/auth/auth.service.ts
Figments: pfft so that's where it kept grabbing `String` instead of `string` from

packages/client/src/app/components/site-sidenav/notifications/notif-item/notif-item.component.ts
Figments: i'll have to work on this in a future update, as currently it breaks a few things

packages/client/src/app/components/site-sidenav/notifications/notification-select.model.ts
Figments: yeah, i think i'll start doing some code cleanup like this for 0.7 alongside the transition to angular universal

pingzing (approved):

PR #344, opened 2021-01-24
fix/cover-art-uploading

### LINKED ISSUE #338: Cover art uploading uses the old method of checking the access token, which can lead to failures to upload
From Figments on Discord:
simple explanation
when you log in, we pull your user info from the server (basically almost everything except your password)
along with your stuff, there's an access token
access tokens expire every hour, but they're automatically regenerated if you have a refresh token cookie set in your browser
but the way tokens are set changed with 0.6
so the coverart upload thingy is still using the old way of looking up your access token
which isn't gonna work anymore

Figments:
Fixes:
- Cover art should now use the correct access token (Closes #338)


SigmasonicX (approved):
LGTM

PR #356, opened 2021-01-27
Consolidate all mini HTTP services into a single NetworkService

pingzing:


packages/client/src/app/pages/my-stuff/section-item/section-item.component.ts
```diff
@@ -41,25 +40,6 @@ export class SectionItemComponent implements OnInit, OnChanges {
         });
     }
 
-    ngOnChanges(): void {
-        // First, clear the contents of the Body and Author's Note. This is because
```
pingzing: Little bit of Quill cleaning.

packages/client/src/app/services/user/notifications.service.ts
```diff
@@ -19,7 +18,7 @@ export class NotificationsService {
 
     private url = `/api/notifications`;
 
-    constructor(private http: HttpClient, private zone: NgZone) {
+    constructor(private networkService: NetworkService, private zone: NgZone) {
         this.currentUserSubscription = this.currentUser$.subscribe(x => {
             this.currentUser = x;
```
pingzing: I left this alive, because it maintains a bit of state, and can serve as a wrapper around its network calls--I can see it doing more stuff in the future, too.

packages/client/src/app/services/utility/search.service.ts
```diff
@@ -1,62 +1,39 @@
...
 
 @Injectable({
-  providedIn: 'root'
+    providedIn: 'root'
 })
 export class SearchService {
-  private url: string = `/api/search`
+    workResults: Observable<PaginateResult<ContentModel>>;
+    blogResults: Observable<PaginateResult<BlogsContentModel>>;
+    userResults: Observable<PaginateResult<User>>;
```
pingzing: I left this alive for reasons similar to NotificationsService even though it doesn't have any state _yet_.

packages/client/src/app/services/network.service.ts
```diff
@@ -0,0 +1,1450 @@
...
+
+    //#region ---COVER ART & AVATARS---
+
+    /**
+     * Uses the given uploader to an image up to the server. Attempts to cast
+     * the server's response into a T and return it.
+     * 
+     * @param uploader The file uploader, prefilled with the URL and instructions for uploading the image.
+     */
+    public changeImage<T extends FrontendUser | ContentModel>(uploader: FileUploader): Observable<T> {
```
pingzing, line 740: Finally get to consolidate this method into one place instead of repeating it three different times \o/

Figments (approved):

PR #367, opened 2021-02-07
cleanup/server-rearrangements

Figments:
Begins work on rearranging the API server to be much more streamlined and readable.


packages/server/src/main.ts
```diff
@@ -25,32 +24,26 @@ async function bootstrap() {
...
     app.use(urlencoded({ limit: '50mb', extended: true }));
-    app.enableCors({
-        origin: process.env.ORIGIN,
-        credentials: true,
-        allowedHeaders: ['Accept', 'Authorization', 'Content-Type', 'Origin'],
-        methods: ['GET', 'PUT', 'POST', 'PATCH', 'DELETE', 'OPTIONS'],
-        optionsSuccessStatus: 204
+    app.use(csurf({ cookie: { sameSite: 'none', secure: true } }));
+    // this is to make sure the XSRF-TOKEN is being set correctly
+    app.use(function (req, res, next) {
```
Figments, line 30: just xsrf now. so long, cors!

README.md
```diff
@@ -37,7 +37,7 @@ Once you've installed and verified that these dependencies are working as expect
 -   Create a file named `.env` at the root of the repository
 -   Copy the contents of `sample.env` to your new `.env`
 -   Edit `.env` to set MONGO_URL=mongodb://localhost:27017
--   Edit `.env` to set JWT_SECRET to an _actual_ secret, such as a plain random string
+-   Edit `.env` to set JWT*SECRET to an \_actual* secret, such as a plain random string
```
pingzing: Pretty sure this isn't what you want either:

image

```suggestion
-   Edit `.env` to set `JWT_SECRET` to an _actual_ secret, such as a plain random string
```

pingzing (commented):
A thought as I'm reading through this: there seems to be a bunch of model duplication with `xxxDTO` objects implementing shared request/response objects, just for the sake of adding Swagger attributes so we can get auto-swagger gen. Can we just add those attributes to the interfaces up in the shared models project? Or if it requires classes, could we just make them be shared classes instead of shared interfaces?

Otherwise, we've got a lot of duplication. It's not the end of the world, because they're yoked to the interface definitions, so there's no risk of things falling out of sync, but it would be nice not to have things repeated.

packages/server/src/app/services/admin/approval-queue.service.ts
```diff
@@ -0,0 +1,35 @@
...
+import { ContentKind, ContentModel } from '@dragonfish/models/content';
+import { ApprovalQueueStore } from '../../db/approval-queue/approval-queue.store';
+import { ContentStore } from '../../db/content';
+import { IApprovalQueue } from '../../shared/admin';
+
+@Injectable()
+export class ApprovalQueueService implements IApprovalQueue {
+    private readonly logger: Logger = new Logger(ApprovalQueueService.name);
+
+    constructor(private approvalQueueService: ApprovalQueueStore, private contentService: ContentStore) {}
```
pingzing: `approvalQueueService` should be renamed to `approvalQueueStore`. Ditto `contentService`.

oh god how many of these have I missed until now

packages/server/src/app/services/admin/meta.service.ts
```diff
@@ -0,0 +1,30 @@
...
+import { Injectable, Logger } from '@nestjs/common';
+
+import { UsersStore } from '../../db/users/users.store';
+import { IMeta } from '../../shared/admin';
+
+@Injectable()
+export class MetaService implements IMeta {
+    private readonly logger: Logger = new Logger(MetaService.name);
+
+    constructor(private readonly usersService: UsersStore) {}
```
pingzing: `userService` to `usersStore`

packages/server/src/app/services/auth/auth.service.ts
```diff
@@ -0,0 +1,94 @@
...
+import { User, FrontendUser, CreateUser, AuditSession } from '@dragonfish/models/users';
+import { UsersStore } from '../../db/users/users.store';
+import { JwtPayload } from '@dragonfish/models/auth';
+import { IAuth } from '../../shared/auth';
+
+@Injectable()
+export class AuthService implements IAuth {
+    private readonly logger: Logger = new Logger(AuthService.name);
+
+    constructor(private readonly usersService: UsersStore, private readonly jwtService: JwtService) {}
```
pingzing: `userService` to `usersStore`

packages/server/src/app/services/auth/auth.service.ts
```diff
@@ -0,0 +1,94 @@
...
+                } else {
+                    throw new UnauthorizedException('Either your email or password is invalid.');
+                }
+            } catch (err) {
+                throw new InternalServerErrorException(
+                    'Something went wrong logging you in. Try again in a little bit.',
+                );
+            }
+        } else {
+            throw new NotFoundException("Looks like you don't exist yet. Why not try signing up?");
```
pingzing, line 38: You were asking about error logging--places like this should all be logged (unless we get exception logging by default, in which case that's good enough. Unless there's more data we'd like to record about the error that an exception alone won't capture)

packages/server/src/app/services/auth/user.service.ts
```diff
@@ -0,0 +1,105 @@
...
+    ChangeProfile,
+    UpdateTagline,
+} from '@dragonfish/models/users';
+import { IUser } from '../../shared/auth';
+
+@Injectable()
+export class UserService implements IUser {
+    private readonly logger: Logger = new Logger(UserService.name);
+
+    constructor(private readonly usersService: UsersStore) {}
```
pingzing: service to store

packages/server/src/app/services/auth/user.service.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    async getOneUser(userId: string): Promise<FrontendUser> {
+        return await this.usersService.getOneUser(userId);
+    }
+
+    async changeUsername(jwtPayload: JwtPayload, changeUsernameRequest: ChangeUsername): Promise<FrontendUser> {
+        const potentialUser = await this.usersService.findOneById(jwtPayload.sub);
+        if (!potentialUser) {
+            // This happening is _super_ fishy. Either a well-meaning tinkerer playing with the API, or something malicious.
```
pingzing, line 35: Like this, for example: this is a great place to log a warning _AND_ throw, because this is a _weird_ and potentially alarming thing.

packages/server/src/app/services/content/collections.service.ts
```diff
@@ -0,0 +1,62 @@
...
+        return await this.collections.getOneCollection(user, collectionId);
+    }
+
+    async getAllCollections(user: JwtPayload, pageNum: number): Promise<PaginateResult<Collection>> {
+        return await this.collections.getAllCollections(user, pageNum);
+    }
+
+    async getPublicCollection(userId: string, collectionId: string): Promise<Collection> {
+        return await this.collections.getOnePublicCollection(userId, collectionId);
+    }
```
pingzing, line 25: And a service like this is an instance where the distinction between a Store and Service is kinda _eeeeehhhhh_, you know? In some places, you want that extra layer. Here, it's kind of unnecessary.

packages/server/src/app/services/index.ts
```diff
@@ -0,0 +1,80 @@
...
+import { AuthServices } from './auth';
+import { ContentServices } from './content';
+import { ImagesServices } from './images';
+import { SearchServices } from './search';
+
+export const InterfaceProviders = [
+    {
+        provide: 'IApprovalQueue',
+        useClass: AdminServices.find((x) => {
+            return x.name === 'ApprovalQueueService';
```
pingzing, line 11: hm. I like the way this is decoupled from things, but the hard reliance on a simple string is gonna be annoying when we want to rename things. I don't suppose TypeScript has something like C#'s `nameof(SomeClassHere)`? It'd be lovely if we could do something like `x.name === nameof(ApprovalQueueService)` (where `ApprovalQueueService` here is a _type_ rather than a string), and get automatic refactor support, instead of runtime explosion

packages/server/src/app/services/search/search.service.ts
```diff
@@ -1,18 +1,18 @@
...
+import { ContentStore } from '../../db/content';
 import { User } from '@dragonfish/models/users';
-import { BlogsContentModel, ContentKind, ContentModel } from '@dragonfish/models/content';
+import { ContentKind, ContentModel } from '@dragonfish/models/content';
 
 @Injectable()
-export class SearchService {
-    constructor(private readonly usersService: UsersService, private readonly contentService: ContentService) {}
+export class SearchService implements ISearch {
+    constructor(private readonly usersService: UsersStore, private readonly contentService: ContentStore) {}
```
pingzing: service to store

packages/server/src/app/shared/admin/approval-queue.interface.ts
```diff
@@ -0,0 +1,56 @@
+import { PaginateResult } from 'mongoose';
+import { ApprovalQueue } from '@dragonfish/models/approval-queue';
+import { ContentKind, ContentModel } from '@dragonfish/models/content';
+import { JwtPayload } from '@dragonfish/models/auth';
+
+/**
+ * ## The Approval Queue Interface
+ *
+ * Implement this interface to be able to serve data to associated routes.
+ */
```
pingzing: Nitpick: Describe what this service does rather than what the interface is for. The people reading this docstring will be in a controller trying to figure out what an `IApprovalQueue` _does_. They don't really care if it's an interface or a class or whatever.

packages/server/src/app/shared/admin/meta.interface.ts
```diff
@@ -0,0 +1,30 @@
+import { FrontPageStats } from '@dragonfish/models/stats';
+import { FrontendUser } from '@dragonfish/models/users';
+
+/**
+ * ## The Meta Interface
+ *
+ * Implement this interface to be able to serve data to associated routes.
+ */
```
pingzing: Ditto the above nitpick

packages/server/src/app/shared/auth/auth.interface.ts
```diff
@@ -0,0 +1,84 @@
+import { JwtPayload } from '@dragonfish/models/auth';
+import { User, FrontendUser, CreateUser, AuditSession } from '@dragonfish/models/users';
+
+/**
+ * ## The Authorization Interface
+ *
+ * Implement this interface to be able to serve data to associated routes.
+ */
```
pingzing: you see where I'm going with this

packages/server/src/app/util/decorators/user.decorator.ts
```diff
@@ -0,0 +1,6 @@
+import { JwtPayload } from '@dragonfish/models/auth';
+import { createParamDecorator } from '@nestjs/common';
+
+export const User = createParamDecorator((_data, req) => {
+    return req.user as JwtPayload;
+});
```
pingzing, line 6: I love this.

packages/server/src/main.ts
```diff
@@ -25,32 +24,32 @@ async function bootstrap() {
...
+        .addTag(`Auth`)
+        .addTag(`User`)
+        .addTag(`Content`)
+        .addTag(`Collections`)
+        .addTag(`Comments`)
+        .addTag(`History`)
+        .addTag(`Messages`)
+        .addTag(`Meta`)
+        .addTag(`Approval Queue`)
+        .addTag(`Search`)
```
pingzing: Worth the trouble to `const`-ify these strings and stick them somewhere central that both the implementations and this call use?

pingzing (changes requested):

packages/server/src/app/util/decorators/user.decorator.ts
Figments: knew you would lmao

packages/server/src/app/services/content/collections.service.ts
Figments: yeah, i was thinking that myself; i can fix it in a later PR

Figments:
> A thought as I'm reading through this: there seems to be a bunch of model duplication with `xxxDTO` objects implementing shared request/response objects, just for the sake of adding Swagger attributes so we can get auto-swagger gen. Can we just add those attributes to the interfaces up in the shared models project? Or if it requires classes, could we just make them be shared classes instead of shared interfaces?
> 
> Otherwise, we've got a lot of duplication. It's not the end of the world, because they're yoked to the interface definitions, so there's no risk of things falling out of sync, but it would be nice not to have things repeated.

We 100% can do that, i just didn't want to go through the trouble of changing them lmao

README.md
Figments: yeah i think prettier keeps messing with this; i'll put it on the ignore list

packages/server/src/app/services/auth/auth.service.ts
Figments: let's go over that in a future PR

packages/server/src/app/services/index.ts
Figments: further exploration of this is required

packages/server/src/main.ts
Figments: yup

pingzing (approved):

PR #382, opened 2021-02-14
Add CI/CD stuff for a digital ocean-based flow, and a rough draft provision script

pingzing:


deploy/provision.sh
```diff
@@ -0,0 +1,32 @@
+# Provisions a machine for usage by the 'dragonfish' backend for Offprint.
```
pingzing, line 1: This file is kind of a work-in-progress, but the idea is that the deployment workflow will invoke it after creating the DO VM. Bonus of it being a separate script: easy to version and modify, and if it ever gets more complex, we can copy the whole `/deploy` folder to the remote machine for provisioning.

Figments (approved):

Issue #424, opened 2021-02-23
Users can't log out, even when auth token expires


SigmasonicX:
Two users at the same time found that pages failed to load for them due to the auth token expiring. They weren't even able to log out due to the invalid auth token. This was resolved by clearing cookies for the site.

This is the first time this happened for both users. Previously, they were logged out as expected when they tried to access a page.

SigmasonicX:
We've seen this with more and more users since this issue was opened.

SigmasonicX:
We've determined that in fact, users can't log out in any scenario, if they logged in after v0.6.1. Clicking the log out button brings the user to the Home page, but otherwise does nothing.

SigmasonicX:
Resolved in dragonfish-next update.

PR #447, opened 2021-03-19
Fix issue where can't log out if don't have refreshToken cookie, and fixes issue where doesn't log out when token expires

SigmasonicX:
Old:
Partially addresses https://github.com/OffprintStudios/dragonfish/issues/424
Even if the user doesn't have a refreshToken, still logs out
Doesn't address issue where nothing happens when the token expires, and the token isn't refreshed

New:
Now has user log out when the token expires
- Can't handle null Observable, so return string instead

### DIFF
--- a/packages/client/src/app/services/network.service.ts
+++ b/packages/client/src/app/services/network.service.ts
@@ -245,7 +245,7 @@ export class NetworkService {
                 catchError((err) => {
                     if (err.status === 403) {
                         // A 403 means that the refreshToken has expired, or we didn't send one up at all, which is Super Suspicious
-                        return of<string>("expired");
+                        return of<string>(null);
                     }
                     return throwError(err);
                 }),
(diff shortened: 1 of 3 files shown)


packages/client/src/app/services/network.service.ts
```diff
@@ -245,7 +245,7 @@ export class NetworkService {
                 catchError((err) => {
                     if (err.status === 403) {
                         // A 403 means that the refreshToken has expired, or we didn't send one up at all, which is Super Suspicious
-                        return null;
+                        return of<string>("expired");
```
pingzing: I think you can actually do `of<string>(null)` here. That way the return signature of the method remains true, and you can keep the null check down below instead of checking for a magic string. My mistake here was that this was supposed to return an `Observable<T>`, and not a raw value. (I thought it'd make an already-returned observable emit whatever I returned. It does not!)

pingzing (changes requested):

Figments (approved):

SigmasonicX:
Was incorporated into dragonfish-next branch.

PR #520, opened 2021-05-08
Reenable notifications for comments

### LINKED ISSUE #355: Users get notifications for their own comments
Currently, users get notifications when they comment on their own stories, which is redundant.

pingzing:
Some quick notes:

 - This adds a new optional field to notifications, `creatorUserId`. It's meant to contain the ID of the user who triggered the notification. We use this to avoid sending notifications of (e.g. comments) to the person who triggered the notification in the first place.
 - Currently, only the plumbing for comments notifications is set up.
 - **This only touches the backend. The frontend is (as far as I know) not hooked up.**

(should) close  #355.

### DIFF
--- a/apps/api/src/app/db/notifications/notifications.service.ts
+++ b/apps/api/src/app/db/notifications/notifications.service.ts
@@ -32,6 +32,7 @@ import { CommentNotificationQueueDocument } from './notificationQueue/comment-no
 import { NewsPostNotificationQueueDocument } from './notificationQueue/news-post-notification-queue.document';
 import { PMThreadNotificationQueueDocument } from './notificationQueue/pm-thread-notification-queue.document';
 import { PMReplyNotificationQueueDocument } from './notificationQueue/pm-reply-notification-queue.document';
+import { Logger } from '@nestjs/common/services/logger.service';
 
 const MAX_NOTIFICATIONS_PER_WAKEUP = 100;
 const MAX_MS_PER_WAKEUP = 100;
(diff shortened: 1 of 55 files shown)


apps/api/src/app/db/notifications/notifications.service.ts
```diff
@@ -329,6 +334,10 @@ export class NotificationsService {
         // Put stale notifications back into the processing queue
         await this.cleanUpStaleNotifications();
 
+        const [sec, ns] = process.hrtime(tickStart);
+        const elapsedMs = (sec * NS_PER_SEC + ns) / NS_PER_MS;
+        this.logger.debug(`NotificationsService wakeup completed in ${elapsedMs}ms. Processed ${processed} notifications. Sleeping...`);
```
pingzing, line 339: This will be kind of noisy, but it's logging to the `debug` sink, so that's what we want.

apps/api/src/app/services/content/comments.service.ts
```diff
@@ -1,21 +1,42 @@
...
 import { ContentComment, CreateComment, EditComment } from '@dragonfish/shared/models/comments';
+import { CreateCommentNotification, NotificationKind } from '@dragonfish/shared/models/notifications';
+import { NotificationsService } from '../../db/notifications/notifications.service';
 
 @Injectable()
 export class CommentsService implements IComments {
-    constructor(private readonly comments: CommentsStore) {}
+    constructor(private readonly comments: CommentsStore,
+        private readonly notificationsService: NotificationsService,
+        @Inject('IContent') private readonly contentService: IContent) { }
```
pingzing, line 15: Yeah! Interfaces! wooo

Figments (approved):

PR #529, opened 2021-05-15
Enhancement/likes and dislikes

Figments:
- moves content view state to new akita-based state management
- moves all ratings-specific functionality to its own collection and store in the database


apps/api/src/app/db/content/content.store.ts
```diff
@@ -108,12 +107,11 @@ export class ContentStore {
...
-            if (authorInfo._id === user.sub) {
-                return doc;
-            } else {
+            if (authorInfo._id !== user.sub) {
                 await this.incrementViewCount(contentId);
-                return doc;
             }
+            await this.history.addOrUpdateHistory(user, contentId);
+            return doc;
         }
```
pingzing: So I recognize that a lot of this is an artifact from not having adjusted from the old design, where Stores had a lot of responsibilities--but the ContentStore should 100% not be touching the HistoryStore--there should be a ContentService (which I believe already exists?) in front of this that fetches from the ContentStore, and updates the user's History in this HistoryStore.

I don't think this will ever get fixed unless we rip of the band-aid and start doing things properly, so I'm going to block this PR on this.

apps/api/src/app/db/ratings/ratings.store.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public async addLike(user: JwtPayload, contentId: string) {
+        const doc = await this.ratings.findOne({ contentId: contentId, userId: user.sub });
+        doc.rating = RatingOption.Liked;
+
+        return doc.save().then(async (doc) => {
+            await this.updateCounts(contentId);
+            return doc;
+        });
```
pingzing: Avoid callbacks, especially when we have `async` available.
```suggestion
        const savedDoc = await doc.save();
        await this.updateCounts(contentId);
        return savedDoc;
```

apps/api/src/app/db/ratings/ratings.store.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public async addDislike(user: JwtPayload, contentId: string) {
+        const doc = await this.ratings.findOne({ contentId: contentId, userId: user.sub });
+        doc.rating = RatingOption.Disliked;
+
+        return doc.save().then(async (doc) => {
+            await this.updateCounts(contentId);
+            return doc;
+        });
```
pingzing:
```suggestion
        const savedDoc = await doc.save();
        await this.updateCounts(contentId);
        return savedDoc;
```

apps/api/src/app/db/ratings/ratings.store.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public async setNoVote(user: JwtPayload, contentId: string) {
+        const doc = await this.ratings.findOne({ contentId: contentId, userId: user.sub });
+        doc.rating = RatingOption.NoVote;
+
+        return doc.save().then(async (doc) => {
+            await this.updateCounts(contentId);
+            return doc;
+        });
```
pingzing:
```suggestion
        const savedDoc = await doc.save();
        await this.updateCounts(contentId);
        return savedDoc;
```

apps/api/src/app/db/ratings/ratings.store.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    private async updateCounts(contentId: string) {
+        const newLikedCount = await this.ratings
+            .countDocuments({ contentId: contentId, rating: RatingOption.Liked });
+        const newDislikedCount = await this.ratings
+            .countDocuments({ contentId: contentId, rating: RatingOption.Disliked });
+        console.log(`New Liked Count: ${newLikedCount}\nNew Disliked Count: ${newDislikedCount}`);
+        await this.contentStore.updateRatingCounts(contentId, newLikedCount, newDislikedCount);
```
pingzing: Nnnngaahh. This really makes me nervous. It couples the implementation of the `contentStore` and this `ratingStore` really tightly together.

I think that, instead, there should be a `ratingService` in front of this that's responsible for 1) calling `addLike`/`addDislike`/`setNoVote` in this Store, and 2) updating the rating counts over in `Content`, and then the this store could _JUST_ update and return its own documents.

apps/api/src/app/db/ratings/ratings.store.ts
```diff
@@ -0,0 +1,70 @@
...
+            return doc;
+        });
+    }
+
+    private async updateCounts(contentId: string) {
+        const newLikedCount = await this.ratings
+            .countDocuments({ contentId: contentId, rating: RatingOption.Liked });
+        const newDislikedCount = await this.ratings
+            .countDocuments({ contentId: contentId, rating: RatingOption.Disliked });
+        console.log(`New Liked Count: ${newLikedCount}\nNew Disliked Count: ${newDislikedCount}`);
```
pingzing: Stray `console.log()`, if we're logging, we should use the `Logger`.

libs/client/repository/src/lib/content-view/content-view.service.ts
```diff
@@ -0,0 +1,139 @@
...
+    @SelectSnapshot(UserState.currUser) private currUser: FrontendUser;
+
+    constructor(
+        private contentView: ContentViewStore,
+        private viewQuery: ContentViewQuery,
+        private network: DragonfishNetworkService,
+        private alerts: AlertsService,
+    ) {}
+
+    public fetchContent(contentId: string, kind: ContentKind) {
```
pingzing: Can we get a type annotation on the return value here? It's tough to tell what it's returning without reading the code.

pingzing (changes requested):
A small handful of style change requests, and two larger, structural change requests in the backend. 
I'd really like it if we could remove a bunch of the dependencies in the ContentStore, but I recognize that there's a bunch of old design decisions in there, and we can't do everything at once. 

That said, I think we can avoid compounding the problem here with the new Ratings collection.

libs/api/database/src/lib/content/stores/content.store.ts
```diff
@@ -1,73 +1,87 @@
...
-        private readonly notificationsService: NotificationsService,
-        private readonly history: ReadingHistoryStore,
+        private readonly notifications: NotificationsService,
+        private readonly queue: ApprovalQueueStore,
     ) {}
 
+    //#region ---FETCHING---
+
     /**
      * Fetches one unpublished item from the content collection via ID and ContentKind.
```
pingzing: nit: I think this comment lies--this just fetches the first document it finds, published or unpublished.

libs/api/database/src/lib/content/stores/content.store.ts
```diff
@@ -131,47 +118,44 @@ export class ContentStore {
     }
 
     /**
-     * Fetches all published documents based on kind, limited by page number.
+     * Fetches a section by ID. Performs an extra check to only fetch a published section.
      *
-     * @param pageNum The current page
-     * @param kinds The kind of document to fetch
+     * @param sectionId The second ID
```
pingzing: nit: typo, should be "section"

libs/api/database/src/lib/content/stores/content.store.ts
```diff
@@ -131,47 +118,44 @@ export class ContentStore {
     }
 
     /**
-     * Fetches all published documents based on kind, limited by page number.
+     * Fetches a section by ID. Performs an extra check to only fetch a published section.
```
pingzing: nit: this makes it sounds like the check is always performed, when it's an argument

libs/api/database/src/lib/content/stores/content.store.ts
```diff
@@ -1,73 +1,87 @@
...
-        private readonly poetryService: PoetryStore,
-        private readonly proseService: ProseStore,
+        @InjectModel('Content') private readonly content: PaginateModel<ContentDocument>,
+        @InjectModel('Sections') private readonly sections: Model<SectionsDocument>,
+        private readonly users: UsersStore,
+        private readonly blogContent: BlogsStore,
+        private readonly newsContent: NewsStore,
+        private readonly proseContent: ProseStore,
+        private readonly poetryContent: PoetryStore,
         private readonly sectionsService: SectionsStore,
```
pingzing: super small nit: this variable should probably be called `sections` or `sectionsStore` to fit the new theme

libs/api/database/src/lib/content/stores/content.store.ts
```diff
@@ -278,36 +262,84 @@ export class ContentStore {
...
-        };
-        this.determineContentFilter(paginateQuery, filter);
+        });
+
+        if (thisContent.kind !== ContentKind.PoetryContent && thisContent.stats.words < 750) {
+            throw new BadRequestException(`Content that isn't poetry needs to have a minimum word-count of 750. `);
+        }
+
+        await this.queue.addOneWork(contentId);
+        return await this.pendingWork(contentId, user.sub);
```
pingzing, line 287: This (and the below functions) extra bit of queue management is the kind of "not-quite strictly CRUD database specific" behavior that's riiight on the edge of making me say "this piece of logic belongs in a service". I think it's okay as it is, but it feels like something worth keeping an eye on.

libs/api/database/src/lib/notifications/notificationQueue/blog-notification-queue.document.ts
```diff
@@ -2,5 +2,5 @@ import { BlogNotificationInfo } from '@dragonfish/shared/models/notifications/bl
 import { NotificationQueueDocument } from '../notification-queue.schema';
 
 export interface BlogNotificationQueueDocument extends NotificationQueueDocument, BlogNotificationInfo {
-    readonly _id?: string;
+    readonly _id: string;
```
pingzing, line 5: not having these caused us compiler errors in the past. make sure this compiles.

pingzing (dismissed):

pingzing (approved):

PR #533, opened 2021-05-17
Draft PR for tags

SigmasonicX:
The process for adding a tag is to go to the Create Works page, which now has a Create Tag section, type in a tag, and click Create

### DIFF
--- a/apps/api/src/app/db/content/prose/fandom-tag.document.ts
+++ b/apps/api/src/app/db/content/prose/fandom-tag.document.ts
@@ -1,0 +1,6 @@
+import { FandomTag } from '@dragonfish/shared/models/content';
+import { Document } from 'mongoose';
+
+export interface FandomTagDocument extends FandomTag, Document {
+    readonly _id: string;
+}
--- a/apps/api/src/app/db/content/prose/fandom-tag.store.ts
+++ b/apps/api/src/app/db/content/prose/fandom-tag.store.ts
@@ -1,0 +1,31 @@
+import { CreateFandomTag } from '@dragonfish/shared/models/content';
+import { Injectable } from '@nestjs/common';
+import { InjectModel } from '@nestjs/mongoose';
+import { PaginateModel } from 'mongoose';
+import { FandomTagDocument } from './fandom-tag.document';
+
+@Injectable()
+export class FandomTagStore {
+    constructor(
+        @InjectModel('FandomTag')
+        private readonly fandomTagModel: PaginateModel<FandomTagDocument>
+    ) {}
+
+    /**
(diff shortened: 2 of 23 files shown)


apps/api/src/app/db/content/prose/fandom-tag.document.ts
```diff
@@ -0,0 +1,6 @@
+import { FandomTag } from '@dragonfish/shared/models/content';
+import { Document } from 'mongoose';
+
+export interface FandomTagDocument extends FandomTag, Document {
+    readonly _id: string;
+}
```
pingzing, line 6: Just thinking aloud here: the name of this alone makes me wonder how much we want to specialize tags. Should have have a base tag interface that both `FandomTag` and `UserTag` inherit from? Should `UserTag` and `FandomTag` even be different? Worth thinking about.

apps/api/src/app/db/content/prose/fandom-tag.store.ts
```diff
@@ -0,0 +1,31 @@
...
+        const newFandomTag = new this.fandomTagModel({
+            name: fandomTagInfo.name,
+            desc: fandomTagInfo.desc,
+            parent: fandomTagInfo.parent,
+            children: fandomTagInfo.children,
+        });
+
+        const savedFandomTag: FandomTagDocument = await newFandomTag.save();
+
+        return savedFandomTag;
```
pingzing, line 29: nit: this can just be `return await newFandomTag.save();`

apps/api/src/app/services/content/fandom-tag.service.ts
```diff
@@ -0,0 +1,17 @@
...
+import { BadRequestException, Injectable, Logger } from '@nestjs/common';
+import { FandomTagStore } from '../../db/content/prose/fandom-tag.store';
+import { IFandomTag } from '../../shared/content/fandom-tag.interface';
+
+@Injectable()
+export class FandomTagService implements IFandomTag {
+    private readonly logger: Logger = new Logger(FandomTagService.name);
+
+    constructor(
+        private readonly fandomTag: FandomTagStore
```
pingzing, line 11: nit: this would be clearer if it were `fandomTagStore`.

apps/api/src/app/shared/content/fandom-tag.interface.ts
```diff
@@ -0,0 +1,5 @@
+import { CreateFandomTag, FandomTagModel } from "@dragonfish/shared/models/content";
+
+export interface IFandomTag {
```
pingzing, line 3: So I know that you're following the pattern here, but I don't love the pattern: something named `IFandomTag` makes me think that this is an individual _tag_, not a service (or interface in this case) that _manages_ tags.

PR #549, opened 2021-05-23
remove email from frontend user object

Figments:
addresses a security issue where the frontend user object was mistakingly leaking user emails


SigmasonicX (approved):
LGTM

PR #561, opened 2021-06-03
Fix/frontend comments

Figments:
- migrates comments to own library module
- adds comment state to content view state
- readjusts some interfaces so they're actually implementing what they say they're implementing


libs/client/comments/src/lib/comments.component.ts
```diff
@@ -0,0 +1,53 @@
...
+    styleUrls: ['./comments.component.scss'],
+})
+export class CommentsComponent {
+    @Input() itemId!: string; // The ID of the content
+    @Input() kind!: CommentKind; // The kind of comments thread it is, between all content
+    @Input() pageNum!: number; // The requested page number
+    @Output() emitPageChange = new EventEmitter<number>(); // Emits the current page number
+
+    newCommentForm = new FormGroup({
+        body: new FormControl('', [Validators.required, Validators.minLength(10)]),
```
SigmasonicX, line 19: Did we want the minimum number of characters to be that high? I wouldn't disagree, but just noting this removes comments like "Neat" and "Meh".

SigmasonicX (approved):
LGTM

PR #562, opened 2021-06-04
Feature/lite editor

### LINKED ISSUE #539: Create a new editor for comments and messages
- should support markdown
- should support mentions
- should support emoji

a good idea of what we should aim for is something like this:

Screen Shot 2021-05-18 at 12 03 26 AM

Figments:
- adds a new comment-specific editor (closes #539 )
- ties comments in with content view state


SigmasonicX (approved):
LGTM

PR #565, opened 2021-06-10
Enhancement/even more editor fixes

Figments:
- adds a bubble and floating toolbar on text selection and text entry
- moves all editor icons over to Remix Icons


SigmasonicX:
Also sets it so that clicking on links doesn't open links in editor

PR #573, opened 2021-06-12
Feature/tree tagging

pingzing:
This is done on the backend, but I ran out of steam before I could get it properly hooked up to the the frontend.

It all compiles, and works though. I'll go through the files, and add some notes where relevant.

Looks like I need to clean up some merge issues too.


apps/api/src/app/services/content/tags.service.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (!tagToDelete) {
+            throw new NotFoundException(`No tag with the ID ${id} could be found.`);
+        }
+
+        await this.contentService.removeTagReferences(tagToDelete._id);
+
+        // Then, delete the tag itself
+        const deletedDocument = await this.tagsStore.deleteTag(id);
+        return deletedDocument;
+    }
```
pingzing: This method is a great example of why I wanted a separation between controllers and stores. Now we have a service that can pull in the Store of the thing it's in charge of, as well as a another service. The TagsStore now doesn't need to know a damn thing about Content, and the controller doesn't get tied up with business logic.

libs/api/database/src/lib/content/schemas/blogs-content.schema.ts
```diff
@@ -25,6 +25,10 @@ export class BlogsContentDocument extends Document implements BlogsContentModel
     };
 
     readonly kind: ContentKind;
+
+    @Prop({type: [{type: MongooseSchema.Types.ObjectId, ref: 'TagsDocument'}] })
+    tags: TagsModel[];
```
pingzing: The syntax is a little wacky here, but it should be correct.

libs/api/database/src/lib/content/schemas/poetry-content.schema.ts
```diff
@@ -30,7 +31,7 @@ export class PoetryContentDocument extends Document implements PoetryContent {
         category: { type: String, enum: Object.keys(WorkKind), required: true },
         form: { type: String, enum: Object.keys(PoetryForm), required: true },
         collection: { type: Boolean, default: false },
-        fandoms: { type: [String], default: null },
+        fandoms: { type: [MongooseSchema.Types.ObjectId], default: null, ref: 'TagsDocument' },
```
pingzing: Let's actually have a bit of a discussion about this: as implemented in this PR, Tags can work for Fandoms, User Tags, or Warning tags. Do we want a separate field on each bit of content that's just a subset of the actual Tags field? I'm not really sure. Would love input.

libs/api/database/src/lib/content/stores/content.store.ts
```diff
@@ -509,4 +509,20 @@ export class ContentStore {
...
+    /**
+     * Go through every piece of content, and remove the given `tagId` from 
+     * its `tags` array.
+     * @param tagId The ID of the tag to remove.
+     */
+    public async removeTagReferences(tagId: string): Promise<void> {
+        await this.content.updateMany(
+            {},
+            { $pull: {tags: { _id: tagId }} }
+        );
```
pingzing, line 523: @Figments, please sanity check this logic, you're Mongoosier than I am

libs/api/database/src/lib/content/stores/tags.store.ts
```diff
@@ -0,0 +1,167 @@
...
+    async findOne(id: string): Promise<TagsDocument> {
+        return this.tags.findById(id);
+    }
+
+    /**
+     * Returns the given tag and all its descendants.
+     * @param king
+     * @param tagId
+     */
+    async fetchDescendants(tagId: string): Promise<TagsTree> {
```
pingzing, line 39: A note: getting a list of all _descendants_ will be relatively slow, but I don't expect a tag to ever have too many descendants. Even a few hundred should be fine, especially if we stick a caching layer on the controller, or in the service, or somewhere.

libs/api/database/src/lib/content/stores/tags.store.ts
```diff
@@ -0,0 +1,167 @@
...
+        const root: TagsTree = {
+            ...parentTag.toObject(),
+            children: []
+        };
+        const tree = await this.buildDescendantTree(root);
+
+        return tree;
+    }
+
+    private async buildDescendantTree(tagNode: TagsTree): Promise<TagsTree> {
```
pingzing, line 54: aaaaa recursion, aaaa

Note: If we want to limit the depth-hell on this, we can do a few things: on tag creation we could do something like "if this tag you're creating has more than X parents, scream and say NO". or we could refuse to iterate beyond a depth of X here in our search method. we got options.

libs/api/database/src/lib/content/stores/tags.store.ts
```diff
@@ -0,0 +1,167 @@
...
+        const results: TagsTree[] = await this.tags.aggregate<TagsTree>([
+            { $match: { _id: id } }
+        ])
+        .graphLookup({
+            from: 'tags',
+            startWith: '$_id',
+            connectFromField: '_id',
+            connectToField: 'parent',
+            as: 'children', // Careful: this string should match the name of the property in `TagsTree`.
+            maxDepth: 0,
```
pingzing, line 88: `graphLookup` is some fuckin' voodoo magic. seriously, check out how it works. I'm force-limiting it to a `maxDepth` of 0 and building the child tree by hand, because it gives us back a flat list by default, which ain't all that useful.

But if all we needed was a flat list, it's the tool for the job

libs/api/database/src/lib/content/stores/tags.store.ts
```diff
@@ -0,0 +1,167 @@
...
+     * @param form
+     * @param tagKind
+     */
+    async createTag(form: TagsForm, tagKind: TagKind): Promise<TagsDocument> {
+        if (form.parent) {
+            // Ensure the parent exists and its kind matches
+            const parentTag = await this.tags.findById(form.parent);
+            if (!parentTag) {
+                throw new NotFoundException("Could not find any parent tag with that ID.");
+            }
```
pingzing, line 110: I tried to guard against edge cases here, but I feel really suspicious I'm not thinking of some.

libs/shared/src/lib/models/content/tag-kind.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum TagKind {
+    Fandom = 'Fandom',
+    Warning = 'Warning',
+    User = 'User',
```
pingzing, line 4: Added user tags here

apps/api/src/app/controllers/content/tags.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    @UseGuards(RolesGuard([Roles.User]))
+    @Get('fetch-tags')
+    async fetchTags(@Query('kind') kind: TagKind): Promise<TagsModel[]> {
+        return await this.tagsService.fetchTags(kind);
+    }
+
+    @UseGuards(RolesGuard([Roles.User]))
+    @Get('fetch-descendants')
+    async fetchChildren(@Query('id') id: string): Promise<TagsTree> {
```
SigmasonicX: It's odd that you alternate between "fetch descendants" and "fetch children".

apps/api/src/app/services/content/content.service.ts
```diff
@@ -84,4 +84,8 @@ export class ContentService implements IContent {
             throw new BadRequestException(`Invalid content kind.`);
         }
     }
+
+    public async removeTagReferences(tagId: string): Promise<void> {
```
SigmasonicX, line 88: It's odd to see a remove method in this file without an add method

apps/bettafish/src/app/app.component.ts
```diff
@@ -34,7 +34,7 @@ export class AppComponent implements OnInit, AfterViewInit {
     ) {}
 
     ngOnInit(): void {
-        this.appQuery.theme$.pipe(untilDestroyed(this), delay(600)).subscribe(theme => {
+        this.appQuery.theme$.pipe(untilDestroyed(this)).subscribe(theme => {
```
SigmasonicX, line 37: Figments added this delay for a reason, so I'd rather not remove it in an unrelated PR.

libs/api/database/src/lib/content/schemas/poetry-content.schema.ts
SigmasonicX: Sorry, I'm having trouble parsing your question. Can you give an example?

libs/api/database/src/lib/content/stores/tags.store.ts
SigmasonicX: For the case of someone getting access to a mod account and using it to just add a bunch of child tags, or some other scenario I'm not thinking of, we should perhaps limit it to just three levels. I don't think we have plans to go further than that.

SigmasonicX: This code checks out. At first I thought there was an issue, but I initially missed that the child.children was being set in the loop, and not the child itself. My notes posted below for others to look at if they don't understand.

a
b c
d e | f g

1
Input: a
currentRoot.children[b].children = bDT(b).children
currentRoot.children[c].children = bDT(c).children
return currentRoot

2
Input: b
currentRoot.children[d].children = bDT(d).children
currentRoot.children[e].children = bDT(e).children
return currentRoot

3
Input: d
d has null children
return d

2
Input: b
currentRoot.children[d].children = d.children = null
currentRoot.children[e].children = bDT(e).children
return currentRoot

4
Input: e
e has null children
return e

2
Input: b
currentRoot.children[d].children = d.children = null
currentRoot.children[e].children = e.children = null
return currentRoot // b - d, e

1
Input: a
currentRoot.children[b].children = (b - d, e).children = [d, e]
currentRoot.children[c].children = bDT(c).children
return currentRoot

libs/api/database/src/lib/content/stores/tags.store.ts
SigmasonicX: Nothing I can think of.

apps/api/src/app/services/content/tags.service.ts
Figments: yeah we should really look into separating out concerns more than we do currently. i've been really guilty of that

libs/api/database/src/lib/content/schemas/blogs-content.schema.ts
Figments: looks good to me

libs/api/database/src/lib/content/schemas/poetry-content.schema.ts
Figments: the problem with doing that is how would we separate out which tags are which? we could use the tags' `kind`, but that creates additional overhead on the client side to parse out and arrange the tags in our desired structure, e.g. fandom tags in the top tag row, content warning tags bottom left, etc

libs/api/database/src/lib/content/stores/content.store.ts
Figments: this seems like it'd be INSANELY slow, assuming you're trying to implement the ability for moderators to remove tags altogether

libs/api/database/src/lib/content/stores/tags.store.ts
Figments: lgtm

libs/api/database/src/lib/content/stores/tags.store.ts
Figments: lgtm

libs/api/database/src/lib/content/stores/tags.store.ts
Figments: interesting, that'll be useful in the future

libs/api/database/src/lib/content/stores/tags.store.ts
Figments: i can't think of anything either

libs/api/database/src/lib/content/schemas/poetry-content.schema.ts
pingzing: That feels like a small price to pay in exchange for a much simpler data model in the database and on the backend. Maybe we expose them to the client as three different types (i.e. actually separate interfaces), but I think it makes sense to treat them all as the same thing.

Worst-case scenario, a handful of `switch` statements in the client doesn't feel like a _terrible_ imposition?

libs/api/database/src/lib/content/stores/content.store.ts
pingzing: My understanding is that `updateMany` is implemented with an eye toward performance, and is notably faster than "update all of these things one by one".

In addition, I'm _hoping_ that tag deletion will be an extremely rare event, but something we should still support. If it's a moderator-only action, I feel like it's probably fine if performance sucks a little bit.

The biggest concern is how hard it'll nuke application performance while it's happening--that definitely feels like an issue in single-threaded Node-land.

Maybe we keep this around for now, while our dataset is still tiny, and consider replacing it with something smarter around the end of beta? We definitely _need_ to be able to delete tags, and leaving dangling tag IDs in works seems like a great way to create lots of work for us later =/

SigmasonicX:
Requesting this be merged in now. Currently, this will allow moderators to add parent tags.
Work to be done:
* Determine if storing IDs as ObjectIds rather than strings is fine.
* Finalize how tags should be displayed on Dashboard given the data structure we're using lists the parents of tags and not the children.
* Re-enable the ability to add a child tag to a tag. Make sure to disable adding child tags to child tags.

Figments (dismissed):

libs/client/dashboard/src/lib/components/tags-management/child-tag-item/child-tag-item.component.ts
```diff
@@ -1,19 +1,19 @@
...
 import { ChildTagFormComponent } from '../child-tag-form/child-tag-form.component';
 import { PopupModel } from '@dragonfish/shared/models/util';
 import { PopupComponent } from '@dragonfish/client/ui';
+import { TagsModel } from '@dragonfish/shared/models/content';
 
 @Component({
     selector: 'dragonfish-child-tag-item',
     templateUrl: './child-tag-item.component.html',
 })
 export class ChildTagItemComponent {
```
pingzing, line 14: Should this component exist at all, given that the concept of a child tag doesn't really exist anymore? (At least not as a unique thing--any tag can conceivably be a child OR a parent tag)

libs/client/dashboard/src/lib/pages/tags-management/tags-management.component.ts
```diff
@@ -27,7 +27,12 @@ export class TagsManagementComponent implements OnInit {
     }
 
     addChild(parentId: string) {
-        this.dialog.open(ChildTagFormComponent, { data: { parentId: parentId }});
+        const alertData: PopupModel = {
+            message: 'Currently disabled',
+            confirm: true,
+        };
+        const dialogRef = this.dialog.open(PopupComponent, { data: alertData });
+        // this.dialog.open(ChildTagFormComponent, { data: { parentId: parentId }});
```
pingzing, line 35: What's the commented-out line for?

pingzing (commented):
(I can't actually approve this, as I'm the PR author)

Looks good. I'd recommend--but won't block on-- the following:
 - Changing the ObjectIDs in the Tag schema to use nanoID instead of ObjectID the way we do elsewhere. 
 - I'd also recommend removing the `User` variant from the `TagKind` enum, as Dash and I (and I think you too, Sigma) had a long discussion about how user tags are really a whole different beast.
 - Consider removing (or maybe renaming--I'm not entirely sure what its purpose is) the `child-tag-component`.

libs/client/dashboard/src/lib/pages/tags-management/tags-management.component.ts
SigmasonicX: That triggers the flow for adding a tag as the child of the given tag. I've disabled it for now.

Figments (approved):

PR #584, opened 2021-06-30
Enhancement/better nav

Figments:
- replaces topnav and sidenav with global components
- streamlines nav elements
- adds new settings page
- removes old sidenav elements


SigmasonicX (approved):
LGTM

Issue #585, opened 2021-06-30
Mobile view broken after global navigation update


SigmasonicX:
![image](https://user-images.githubusercontent.com/71996084/123916502-9408d400-d936-11eb-9845-685805e4587a.png)

This is how every page looks on mobile view. There is no way to scroll past the navigation bar.

Figments:
fixed in current main

PR #586, opened 2021-06-30
people can actually use mobile now

Figments:


SigmasonicX (approved):
Approved as a temp fix for https://github.com/OffprintStudios/dragonfish/issues/585

PR #588, opened 2021-07-03
Enhancement/more nav in more places

Figments:
- re-adds hard links to docs pages
- separates registration and login to dedicated page for use in future expansion


SigmasonicX (approved):
Looks good on laptop
Looks good—or doesn't look worse—on mobile.

Issue #627, opened 2021-07-27
Fandom tags Dashboard integration TODO


SigmasonicX:
The following needs to be done to integrate fandom tags into the Dashboard
- Finalize how tags should be displayed on Dashboard given the data structure we're using lists the parents of tags and not the children.
- Re-enable the ability to add a child tag to a tag. Make sure to disable adding child tags to child tags.
- Remove child tag specific code.
- Split off user tags from the Tags enum.
- Improve database speed with indexing.

SigmasonicX:
Resolved by #629

SigmasonicX:
Follows up on #573

PR #629, opened 2021-07-28
Fully integrate fandom tags into Dashboard

SigmasonicX:
Addresses ticket #627

## Description
Fully implements the ability to create and edit fandom tags in the Dashboard

## Changes
- Restores ability to add child tags
- Creates fetchTagsTrees() and sorts tags into TagsTrees
- Has TagsState use TagsTrees instead of TagsModels
- In TagsModel, has parent just be string
- In TagsTree, makes children optional
- Removes User from TagKind, since will use different logic for user tags
- Removes tag MongoDB index on parent, adds on { parent: 1, name: 1 }
- Removes tag accessing methods that we no longer have a clear use case for
- Tag deletion now clears parent from the tag's children
- When updating tag, throws error if it sets its parent to itself; or if the parent is set to another tag and this tag has children, to avoid adding an extra layer
- When editing a child tag, opens the general tag dialog
- Adds ability to set parent on create and edit tags dialog
- Removes ChildTagFormComponent, incorporating child tag functionality into TagFormComponent
- Makes desc optional

## Areas Affected
- [ ] Site Navigation
- [ ] Home
- [ ] Sign In/Out
- [ ] Registration
- [ ] Settings

.
- [ ] Profile
- [ ] My Stuff (main page)
- [ ] Editor
- [ ] Browse
- [ ] Work Card

.
- [ ] Work Page
- [ ] Blog Page
- [ ] Collections
- [ ] Comments
- [ ] Other Pages

.
- [ ] Account Authentication
- [x] Dashboard
- [ ] Mobile


libs/api/database/src/lib/content/stores/tags.store.ts
```diff
@@ -22,6 +22,25 @@ export class TagsStore {
...
+    }
+
+    /**
+     * Returns all tags of a specified kind without parents, sorted alphabetically.
+     * @param kind
+     */
+    async fetchParentTags(kind: TagKind): Promise<TagsDocument[]> {
+        return this.tags.find({
+            kind: kind,
+            parent: null,
```
pingzing: Warning, this might be a bit of a sneaky performance hole, because `parent` isn't indexed. This query will force Mongo to scan every tag in the collection.

...and in fact, now that I'm looking at this, `find({kind:kind})` might be too, for the same reason--neither of these fields have indexes.

...we should really do a bit of profiling here.

libs/api/database/src/lib/content/stores/tags.store.ts
```diff
@@ -22,6 +22,25 @@ export class TagsStore {
         return this.tags.find({ kind: kind }).sort({ name: 1 });
     }
 
+    /**
+     * Returns all tags of a specified kind, sorted by parent (none first) then alphabetically.
+     * @param kind
+     */
+    async fetchTagsSortedByParent(kind: TagKind): Promise<TagsDocument[]> {
+        return this.tags.find({ kind: kind }).sort({ parent: 1, name: 1 });
```
pingzing: You might be able to achieve something similar with Mongo's `graphLookup()`; that can give you a flat list of _all_ tags, but it'll preserve their parent/child relationships. If you want them as a `TagsTree`, you'll have to reconstruct it it a bit, but it might be easier.

libs/client/dashboard/src/lib/pages/tags-management/tags-management.component.ts
```diff
@@ -19,20 +19,15 @@ export class TagsManagementComponent implements OnInit {
...
     }
 
     addChild(parentId: string) {
-        const alertData: PopupModel = {
-            message: 'Currently disabled',
-            confirm: true,
-        };
-        const dialogRef = this.dialog.open(PopupComponent, { data: alertData });
-        // this.dialog.open(ChildTagFormComponent, { data: { parentId: parentId }});
+        this.dialog.open(ChildTagFormComponent, { data: { parentId: parentId }});
```
pingzing: This comes back to what I was talking about in Discord--I don't think we want to pigeonhole ourselves into "one parent, one child" here. I think we should just have an "add tag" form, and that form can have an optional field for "parent". That'll allow us arbitrarily-deep nesting (the whole point of my PR), and if we want to, we can add depth limitations on the backend (or even the frontend, depending on how much it knows about the whole tree).

libs/client/repository/src/lib/dashboard/tags/tags.service.ts
```diff
@@ -27,6 +27,49 @@ export class TagsService {
         );
     }
 
+    public fetchTagsSortedByParent(kind: TagKind): Observable<TagsModel[]> {
+        return this.network.fetchTagsSortedByParent(kind).pipe(
+            tap(value => {
+                var map = new Map<String, TagsTree>();
```
pingzing: two nits:
1) use `let` instead of `var`: `var` has [weird scoping behavior](https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var). I'd call this a general rule of thumb
2) call the variable `tagRoots` or something descriptive

libs/client/dashboard/src/lib/pages/tags-management/tags-management.component.ts
SigmasonicX: In the future, if we want to go to arbitrary fandom tag levels, then I think it's worth redesigning the UI in the future to accommodate our needs. In the meantime, having clear UI that just handles two levels seems like it would be better for now. We need to account for usability with the mods that will be using this.

libs/api/database/src/lib/content/schemas/tags.schema.ts
```diff
@@ -11,10 +11,10 @@ export class TagsDocument extends Document implements TagsModel {
     @Prop({ trim: true, required: true, index: 'text' })
     name: string;
 
-    @Prop({ trim: true, required: true })
-    desc: string;
+    @Prop({ trim: true, default: undefined })
```
pingzing: no need for the `default: undefined`, I believe--pretty sure an optional (i.e., has a `?`) defaults to `null`. If we _do_ define a default value here, I think `null` would be better, as that would ensure it shows up in the JSON payload, and is a teeny bit more explicit.

libs/api/database/src/lib/content/schemas/tags.schema.ts
```diff
@@ -11,10 +11,10 @@ export class TagsDocument extends Document implements TagsModel {
...
+    desc?: string;
 
-    @Prop({ type: String, default: undefined, ref: 'TagsDocument', index: true })
+    @Prop({ type: String, default: undefined, ref: 'TagsDocument' })
```
pingzing: ditto `null` stuff here.

libs/api/database/src/lib/content/schemas/tags.schema.ts
```diff
@@ -28,3 +28,4 @@ export class TagsDocument extends Document implements TagsModel {
 }
 
 export const TagsSchema = SchemaFactory.createForClass(TagsDocument);
+TagsSchema.index({ parent: 1, name: 1 })
```
pingzing, line 34: nit: I almost want to say that we should have a comment mentioning this index up at the top of the file, because a) this is the first time we've used a compound index and b) indices usually get declared inline, so this being down here will be a bit of a surprise.

libs/client/dashboard/src/lib/components/tags-management/tag-form/tag-form.component.ts
```diff
@@ -15,25 +15,37 @@ export class TagFormComponent implements OnInit {
 
     tagForm = new FormGroup({
         name: new FormControl('', [Validators.required]),
-        desc: new FormControl('', [Validators.required, Validators.maxLength(100)]),
+        desc: new FormControl('', [Validators.maxLength(100)]),
```
pingzing: nit: Only 100 feels pretty restrictive, and I think text limits are kind of a bad habit we have all over the codebase. We should bump this, and enforce length limitations on the view side, rather than the data side.

libs/client/repository/src/lib/dashboard/tags/tags.service.ts
```diff
@@ -39,8 +46,8 @@ export class TagsService {
 
     public createTag(kind: TagKind, form: TagsForm): Observable<TagsModel> {
         return this.network.createTag(kind, form).pipe(
-            tap(value => {
-                this.tagsStore.add(value);
+            tap(newTag => {
+                this.fetchTagsTrees(TagKind.Fandom).subscribe();
```
pingzing, line 50: Not super crazy about all these methods having a side-effect that updates the local tag cache, but I don't have a better solution off the top of my head, either.

pingzing (dismissed):
A handful of nits, but nothing I'll block over.

libs/client/dashboard/src/lib/components/tags-management/tag-form/tag-form.component.ts
```diff
@@ -67,8 +67,8 @@ export class TagFormComponent implements OnInit {
         const parentVal = (this.fields.parent.value === this.NO_PARENT)? null : this.fields.parent.value;
 
         const form: TagsForm = {
-            name: this.fields.name.value,
-            desc: this.fields.desc.value,
+            name: (this.fields.name.value as string).trim(),
+            desc: (this.fields.desc.value as string).trim(),
```
pingzing, line 71: minor heebie-jeebies here--if `this.fields.desc.value` can be null, this will crash. I _think_ we're saved by Angular enforcing that the fields of a form default to an empty string instead of null, but not 100% sure that that's always the case.

pingzing (approved):

PR #644, opened 2021-08-10
Feature/pseudonyms

Figments:
## Description

Begins work on the new Accounts and Pseudonyms system. Nothing is hooked up yet, but migration should be functioning correctly.


SigmasonicX (approved):
LGTM

Issue #636, opened 2021-08-04
Create fandom tag pages


SigmasonicX:
Split from #352

When click fandom tag, user is brought to page that displays the tag's hierarchy, the description, and all works with that tag.

Issue #637, opened 2021-08-04
Add advanced searching, which includes implementing searching by tag, and multiple tags


SigmasonicX:
Split from #352

Users should be able to search by a single tag, and multiple tags, combined with other search terms.
When searching a parent tag, you get results for all child tags too. Users have the option to only look for the parent tag and not its children.

- [x] For all pages that fetch multiple content, figure out the flow and see where they can be consolidated.
- [x] Consolidate those methods and classes. Call it fetch-multiple-content.store. This could be its own PR.
- [x] Split Search off into its own page. Remove Search from Browse. Add Search to side menu (bottom on mobile). This could be its own PR.
- [x] On Search page, add "Advanced Options" expansion, and under there, add options to select what results get back from among works, poems, blogs, news, and authors. All are pre-checked. Unchecking then searching will remove those from results.
- [x] Under Advanced Options, option to search by content name AND by author (separate text field). Can search just by author; can still search by author with top field.
- [x] Under Advanced Options, add option to search by original / fanwork. Can search just by this.
- [x] Under Advanced Options, add option to search by genre. Can search just by this.
- [x] For genre search, add ability to search by multiple genre. Can specify if all are required, or if any.
- [x] Under Advanced Options, add option to search by a single tag. Can search just by this.
- [x] For tag search, add checkmark for if tag's children are included, checked by default. Search includes children by default, won't if unchecked.
- [x] Add this same feature to tag pages. With code consolidation, should use same methods as Search in the end.
- [x] For tag search, add ability to search by multiple tags. Can specify if all are required, or if any.

PR #646, opened 2021-08-16
Implements fandom tag pages and fixes Dashboard issues

SigmasonicX:
Addresses ticket #636 and #649

## Description
When you click fandom tag, you go to page that describes the tag and lists stories with that tag.

## Changes
- Creates page for each fandom tag. This includes a link to its parent if available, the description, links to its children if available, and all works with this fandom tag. This does not include all works with its children as fandom tags.
- Changes the tag listings in Dashboard, My Stuff, work cards, and content boxes to have links.
- All changes from #650

## Areas Affected
- [ ] Site Navigation
- [ ] Home
- [ ] Sign In/Out
- [ ] Registration
- [ ] Settings

.
- [ ] Profile
- [x] My Stuff (main page)
- [x] Editor
- [x] Browse
- [x] Work Card

.
- [x] Work Page
- [ ] Blog Page
- [ ] Collections
- [ ] Comments
- [ ] Other Pages

.
- [ ] Account Authentication
- [x] Dashboard
- [ ] Mobile

### DIFF
--- a/apps/api/src/app/controllers/search/search.controller.ts
+++ b/apps/api/src/app/controllers/search/search.controller.ts
@@ -51,4 +51,14 @@ export class SearchController {
     ): Promise<PaginateResult<ContentModel>> {
         return await this.searchService.searchContent(query, pageNum, contentFilter);
     }
+
+    @ApiTags(DragonfishTags.Search)
+    @Get('get-fandom-tag-content')
+    async getFandomTagContent(
+        @Query('tagId') tagId: string,
+        @Query('pageNum') pageNum: number,
+        @Cookies('contentFilter') contentFilter: ContentFilter
+    ): Promise<PaginateResult<ContentModel>> {
+        return await this.searchService.searchFandomTagContent(tagId, pageNum, contentFilter);
+    }
 }
--- a/apps/api/src/app/controllers/search/search.controller.ts
+++ b/apps/api/src/app/controllers/search/search.controller.ts
@@ -53,12 +53,12 @@ export class SearchController {
     }
 
     @ApiTags(DragonfishTags.Search)
-    @Get('get-fandom-tag-content')
-    async getFandomTagContent(
+    @Get('get-content-by-fandom-tag')
+    async getContentByFandomTag(
         @Query('tagId') tagId: string,
         @Query('pageNum') pageNum: number,
         @Cookies('contentFilter') contentFilter: ContentFilter
     ): Promise<PaginateResult<ContentModel>> {
-        return await this.searchService.searchFandomTagContent(tagId, pageNum, contentFilter);
+        return await this.searchService.getContentByFandomTag(tagId, pageNum, contentFilter);
     }
 }
--- a/apps/api/src/app/services/search/search.service.ts
+++ b/apps/api/src/app/services/search/search.service.ts
@@ -12,19 +12,28 @@ import { ContentKind, ContentModel } from '@dragonfish/shared/models/content';
 
 @Injectable()
 export class SearchService implements ISearch {
+    INITIAL_PAGE = 1;
+    INITIAL_MAX_PER_PAGE = 6;
+    MAX_PER_PAGE = 15;
+
     constructor(private readonly usersStore: UsersStore, private readonly contentStore: BrowseStore) {}
 
     async fetchInitialResults(query: string, contentFilter: ContentFilter): Promise<InitialResults> {
         const parsedQuery = `"${sanitizeHtml(query)}"`;
 
         const [initialUsers, initialBlogs, initialContent] = await Promise.all([
-            this.usersStore.findRelatedUsers(parsedQuery, 1, 6),
-            this.contentStore.findRelatedContent(parsedQuery, [ContentKind.BlogContent], 1, 6, contentFilter),
+            this.usersStore.findRelatedUsers(parsedQuery, this.INITIAL_PAGE, this.INITIAL_MAX_PER_PAGE),
+            this.contentStore.findRelatedContent(parsedQuery,
+                [ContentKind.BlogContent],
+                this.INITIAL_PAGE,
+                this.INITIAL_MAX_PER_PAGE,
+                contentFilter
+            ),
             this.contentStore.findRelatedContent(
                 parsedQuery,
                 [ContentKind.PoetryContent, ContentKind.ProseContent],
-                1,
-                6,
+                this.INITIAL_PAGE,
+                this.INITIAL_MAX_PER_PAGE,
                 contentFilter
             ),
         ]);
@@ -39,7 +48,7 @@ export class SearchService implements ISearch {
 
     async searchUsers(query: string, pageNum: number): Promise<PaginateResult<User>> {
         const parsedQuery = `"${sanitizeHtml(query)}"`;
-        return await this.usersStore.findRelatedUsers(parsedQuery, pageNum, 15);
+        return await this.usersStore.findRelatedUsers(parsedQuery, pageNum, this.MAX_PER_PAGE);
     }
 
     async searchBlogs(
@@ -52,7 +61,7 @@ export class SearchService implements ISearch {
             parsedQuery,
             [ContentKind.BlogContent],
             pageNum,
-            15,
+            this.MAX_PER_PAGE,
             contentFilter
         );
     }
@@ -67,8 +76,22 @@ export class SearchService implements ISearch {
             parsedQuery,
             [ContentKind.PoetryContent, ContentKind.ProseContent],
             pageNum,
-            15,
+            this.MAX_PER_PAGE,
             contentFilter
         );
+    }
+
+    async searchFandomTagContent(
+        tagId: string,
+        pageNum: number,
+        contentFilter: ContentFilter
+    ): Promise<PaginateResult<ContentModel>> {
+        return await this.contentStore.findFandomTagContent(
+            tagId,
+            [ContentKind.PoetryContent, ContentKind.ProseContent],
+            pageNum,
+            this.MAX_PER_PAGE,
+            contentFilter
+        )
     }
 }
--- a/apps/api/src/app/services/search/search.service.ts
+++ b/apps/api/src/app/services/search/search.service.ts
@@ -12,9 +12,9 @@ import { ContentKind, ContentModel } from '@dragonfish/shared/models/content';
 
 @Injectable()
 export class SearchService implements ISearch {
-    INITIAL_PAGE = 1;
-    INITIAL_MAX_PER_PAGE = 6;
-    MAX_PER_PAGE = 15;
+    readonly INITIAL_PAGE = 1;
+    readonly INITIAL_MAX_PER_PAGE = 6;
+    readonly MAX_PER_PAGE = 15;
 
     constructor(private readonly usersStore: UsersStore, private readonly contentStore: BrowseStore) {}
 
@@ -81,12 +81,12 @@ export class SearchService implements ISearch {
         );
     }
 
-    async searchFandomTagContent(
+    async getContentByFandomTag(
         tagId: string,
         pageNum: number,
         contentFilter: ContentFilter
     ): Promise<PaginateResult<ContentModel>> {
-        return await this.contentStore.findFandomTagContent(
+        return await this.contentStore.getContentByFandomTag(
             tagId,
             [ContentKind.PoetryContent, ContentKind.ProseContent],
             pageNum,
--- a/apps/bettafish/src/app/app-routing.module.ts
+++ b/apps/bettafish/src/app/app-routing.module.ts
@@ -22,6 +22,7 @@ import { Resolvers } from './resolvers';
 /* Util */
 import { Roles } from '@dragonfish/shared/models/users';
 import { AuthGuard } from '@dragonfish/client/repository/session/services';
+import { TagRoutes } from './pages/tags';
 
 const routes: Routes = [
     ...HomeRoutes,
@@ -33,6 +34,7 @@ const routes: Routes = [
     ...ContentViewRoutes,
     ...SettingsRoutes,
     ...RegistrationRoutes,
+    ...TagRoutes,
     {
         path: 'my-stuff',
         canLoad: [AuthGuard],
--- a/apps/bettafish/src/app/components/content/content-box/content-box.component.html
+++ b/apps/bettafish/src/app/components/content/content-box/content-box.component.html
@@ -60,8 +60,17 @@
                             <div class="tag">{{ $any(content.currContent.meta).genres | separateGenres }}</div>
                         </div>
                         <ng-container *ngIf="content.currContent.tags !== null && content.currContent.tags.length > 0">
-                            <div class="flex flex-row items-center uppercase text-sm">
-                                <span class="tag tags">{{ content.currContent.tags | displayTags }}</span>
+                            <div class="content-tags flex flex-row items-center uppercase text-sm">
+                                <span class="tag tags">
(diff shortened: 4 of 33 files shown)


apps/api/src/app/controllers/search/search.controller.ts
```diff
@@ -51,4 +51,14 @@ export class SearchController {
...
+
+    @ApiTags(DragonfishTags.Search)
+    @Get('get-fandom-tag-content')
+    async getFandomTagContent(
+        @Query('tagId') tagId: string,
+        @Query('pageNum') pageNum: number,
+        @Cookies('contentFilter') contentFilter: ContentFilter
+    ): Promise<PaginateResult<ContentModel>> {
+        return await this.searchService.searchFandomTagContent(tagId, pageNum, contentFilter);
+    }
```
SigmasonicX, line 63: Added method to get fandom tag content as part of search; in the future, this will be consolidated into a more advanced general search method, since #637 will require being able to search by tag anyway.

apps/api/src/app/services/search/search.service.ts
```diff
@@ -12,19 +12,28 @@ import { ContentKind, ContentModel } from '@dragonfish/shared/models/content';
 
 @Injectable()
 export class SearchService implements ISearch {
+    INITIAL_PAGE = 1;
+    INITIAL_MAX_PER_PAGE = 6;
+    MAX_PER_PAGE = 15;
+
```
SigmasonicX, line 18: As these numbers are repeated often throughout this class, I made them fields.

apps/bettafish/src/app/app-routing.module.ts
```diff
@@ -33,6 +34,7 @@ const routes: Routes = [
     ...ContentViewRoutes,
     ...SettingsRoutes,
     ...RegistrationRoutes,
+    ...TagRoutes,
```
SigmasonicX, line 37: To support going to the tag pages.

apps/bettafish/src/app/components/content/content-box/content-box.component.html
```diff
@@ -60,8 +60,18 @@ <h2 class='text-white text-4xl font-medium hover:cursor-pointer'>
...
+                                    <ng-container *ngFor="let tag of content.currContent.tags; let i = index">
+                                        <a 
+                                           [routerLink]="['/tag', tag._id, tag.name | slugify]"
+                                           [innerHtml]="tag | displayTags | safeHtml"
+                                        ></a>
+                                        <ng-container *ngIf="i < content.currContent.tags.length - 1">
+                                            , 
+                                        </ng-container>
+                                    </ng-container>
+                                </span>
```
SigmasonicX, line 74: Adds links to tags, and handles HTML conversion like \&amp; -> &

apps/bettafish/src/app/components/content/content-box/content-box.component.scss
```diff
@@ -37,4 +37,13 @@ div.content-box {
...
     }
+    div.content-tags {
+        a {
+            cursor: pointer;
+            color: whitesmoke;
+            &:hover {
+                background: var(--accent-dark);
+            }
+        }
+    }
```
SigmasonicX, line 48: Makes tags visible as links

apps/bettafish/src/app/pages/content-views/news-page/news-page.component.ts
```diff
@@ -26,7 +26,7 @@ export class NewsPageComponent implements OnInit {
     ) {}
 
     ngOnInit(): void {
-        combineLatest(this.viewState.currContent$, this.route.queryParamMap)
+        combineLatest([this.viewState.currContent$, this.route.queryParamMap])
```
SigmasonicX, line 29: Handles warning by converting parameters to array; separate parameters in this method is deprecated

apps/bettafish/src/app/pages/tags/tag-page.component.html
```diff
@@ -0,0 +1,56 @@
+<dragonfish-topbar></dragonfish-topbar>
```
SigmasonicX: This and the TS file below are the main additions

apps/bettafish/src/material.scss
```diff
@@ -86,35 +86,6 @@ button.mat-menu-item {
...
-    font-weight: 500 !important;
-}
-
-/* Mat Expansion Panel */
-.mat-expansion-panel {
-    background: var(--background) !important;
-    color: var(--text-color) !important;
-    border: 1px solid var(--borders) !important;
-}
-
```
SigmasonicX, line 117: Removes code that is overwritten right below it.

libs/api/database/src/lib/content/schemas/content.schema.ts
```diff
@@ -79,7 +79,7 @@ export class ContentDocument extends Document implements ContentModel {
             select: '_id name desc parent kind createdAt updatedAt',
         },
     }] })
-    tags?: TagsModel[];
+    tags?: string[] | TagsModel[];
```
SigmasonicX, line 82: Needed to check tags with string queries

libs/api/database/src/lib/content/stores/tags.store.ts
```diff
@@ -153,10 +172,20 @@ export class TagsStore {
...
+                throw new BadRequestException("There is already a child tag of this parent with this name.");
+            }
+        }
+        // If parent tag, check that name not already in use by another parent tag
+        // Don't check for duplicates with children
+        else {
+            const tagWithSameName = await this.findParentTagByName(newTag.name)
+            if (tagWithSameName) {
+                throw new BadRequestException("There is already a parent tag with this name.");
+            }
```
SigmasonicX, line 188: Improves name duplication check when creating/editing tags by just checking on the same level

libs/client/dashboard/src/lib/pages/tags-management/tags-management.component.html
```diff
@@ -25,15 +25,17 @@ <h3>Looks like there something missing here...</h3>
             <div class="w-8/12 mx-auto my-8">
                 <mat-accordion>
                     <ng-container *ngFor="let tag of tags">
-                        <mat-expansion-panel>
+                        <mat-expansion-panel [expanded]="(tag._id === openPanelId)? true : false" (opened)="onOpen(tag._id)">
```
SigmasonicX, line 28: Keeps the tag panel open after making changes, as opposed to closing it

libs/client/repository/src/lib/tags/tags.service.ts
```diff
@@ -36,7 +36,9 @@ export class TagsService {
 
     public fetchDescendants(id: string): Observable<TagsTree> {
         return this.network.fetchDescendants(id).pipe(
-            tap(() => {}),
+            tap((tagTree) => {
+                this.tagsStore.set([tagTree]);
+            }),
```
SigmasonicX, line 41: Used in tag pages to get children

libs/shared/src/lib/functions/html-decode.ts
```diff
@@ -0,0 +1,4 @@
+export function htmlDecode(input: string): string {
+    var doc = new DOMParser().parseFromString(input, "text/html");
+    return doc.documentElement.textContent;
+}
```
SigmasonicX, line 4: Method used to handle \&amp; -> & and similar HTML processing outside of HTML files

apps/api/src/app/controllers/search/search.controller.ts
pingzing: nitpick: The name of this method confuses me a little--does it get the content of a fandom tag, or the set of all content _with_ a particular fandom tag? If the latter, maybe `getContentByFandomTag`?

apps/api/src/app/services/search/search.service.ts
pingzing: always a fan. 👍 
nitpick: make 'em const too, to clarify intent

apps/bettafish/src/app/pages/tags/tag-page.component.html
```diff
@@ -1,49 +1,56 @@
...
-                                    </li>
-                                </ng-container>
-                            </ul>
-                        </mat-expansion-panel>
+    <ng-template #notLoading>
+        <ng-container *ngIf="tagsQuery.all$ | async as tags">
+            <ng-container *ngIf="tags.length === 0; else notEmpty">
+                <div class="empty">
+                    <h3>Looks like there something missing here...</h3>
+                    <p>If you think this is an error, contact your system administrator.</p>
```
pingzing, line 14: nitpick: I know this is a standard error (and I think we use it everywhere) but we may want to consider changing this to something more appropriate (i.e. "please let us know in Discord" or something)

apps/bettafish/src/app/pages/tags/tag-page.component.html
```diff
@@ -1,49 +1,56 @@
...
+                </div>
+            </ng-container>
+            <ng-template #notEmpty>
+                <div class="w-11/12 mx-auto my-6">
+                    <h3>
+                        <ng-container *ngIf="tags[0].parent">
+                            <a
+                                [routerLink]="['/tag', tags[0].parent._id, tags[0].parent.name | slugify]"
+                                [innerHtml]="tags[0].parent.name | safeHtml"
```
pingzing, line 23: nitpick: Angular's recommend styles tend to avoid `innerHtml` in favor of `{{ interpolation }}`. I don't think we have a precedent in the codebase, but as a _personal_ preference, I find interpolation a bit more readable, i.e.
```html
<a [routerLink]="['/tag', tags[0].parent._id, tags[0].parent.name | slugify]">
  {{ tags[0].parent.name | safeHtml }}
</a>
```

As an aside, you shouldn't need to run a tag name through `safeHtml`, as tag names shouldn't contain HTML.

...that said, do we have anything preventing a malicious moderator from creating a tag name with HTML tags in it?

apps/bettafish/src/app/pages/tags/tag-page.component.html
```diff
@@ -1,49 +1,56 @@
...
+                            ></a>
+                            &nbsp;— 
+                        </ng-container>
+                        <span [innerHtml]="tags[0].name | safeHtml"></span>
```
pingzing, line 27: nitpick: ditto the `innerHtml` versus interpolation here. etc etc for the rest of the file

apps/bettafish/src/app/pages/tags/tag-page.component.html
```diff
@@ -1,49 +1,56 @@
...
+                            </ng-container>
+                        </ng-container>
                     </ng-container>
-                </mat-accordion>
-            </div>
-        </ng-template>
-    </ng-container>
-</ng-template>
+                    <p>
+                        <span [innerHtml]="tags[0].desc | safeHtml"></span>
```
pingzing, line 42: ditto desc-as-html but oh no do we sanitize that in the backend either

apps/bettafish/src/app/pages/tags/tag-page.component.ts
```diff
@@ -1,25 +1,73 @@
...
+
+    private fetchData(tagId: string, pageNum: number): void {
+        this.tagsService.fetchDescendants(tagId).subscribe((tagsTree) => {
+            if (tagsTree.parent) {
+                setTwoPartTitle(htmlDecode((tagsTree.parent as TagsModel).name + " — " + tagsTree.name));
+            }
+            else {
+                setTwoPartTitle(htmlDecode(tagsTree.name));
+            }
+        });
```
pingzing: We can save some network traffic here. As currently written, we're re-fetching tag data on every page change, but we can cache it, because it's unlikely to change. You'll want to call it once, then just use the cached value, i.e. something like this:

```TypeScript
import { lastValueFrom } from 'rxjs';

// As a class-level variable:
private cachedTags: TagsTree = null;

private async fetchData(tagId: string, pageNum: number): Promise<void> { // now that this is async, remember to make the callback passed to .subscribe() up above async as well)
  if (!this.cachedTags) {
    this.cachedTags = await lastValueFrom(this.tagsService.fetchDescendants(tagId));
    if (this.cachedTags.parent) {
      // etc
    }
    else {
     // etc
    } 
  }
// Though if you wanted, you could turn this into an awaitable promise with lastValueFrom() too!
this.network.searchFandomTagContent(tagId, pageNum).subscribe((results) => {
  this.searchResults = results;
});
}

```

libs/shared/src/lib/functions/html-decode.ts
pingzing: oh good, hopefully we can stop having amps in places where people don't want them. 👍

pingzing (changes requested):
The only change I actually care about is the caching of tags on the `tag-page-component.ts`. Everything else is minor stuff you can choose to ignore if you'd lile.

apps/bettafish/src/app/pages/tags/tag-page.component.html
SigmasonicX: That's how it was before, and it didn't process the HTML correctly

apps/bettafish/src/app/pages/tags/tag-page.component.html
SigmasonicX: I think updating that is for another PR

apps/bettafish/src/app/pages/tags/tag-page.component.html
SigmasonicX: Regarding creating tags with HTML tags in the name, we run sanitize() on the names and descriptions; this is just being doubly safe.

apps/bettafish/src/app/pages/tags/tag-page.component.ts
SigmasonicX: Since the tag tree is already stored in the LocalTagsStore and accessed with TagsQuery, there isn't a need to save it as a field. When clicking around, I noticed some odd behavior that I accounted for in the new code.

pingzing: oh right, I keep forgetting we have a Store layer that already caches for us between pages and the network service...

apps/api/src/app/services/search/search.service.ts
pingzing: oh right, class variables can't be const. I keep forgetting that...

apps/bettafish/src/app/pages/tags/tag-page.component.ts
```diff
@@ -33,24 +33,30 @@ export class TagPageComponent implements OnInit {
...
+        if (this.tagId != tagId) {
+            this.tagId = tagId;
+            this.tagsService.fetchDescendants(tagId).subscribe((tagsTree) => {
+                if (tagsTree.parent) {
+                    setTwoPartTitle(htmlDecode((tagsTree.parent as TagsModel).name + " — " + tagsTree.name));
+                }
+                else {
+                    setTwoPartTitle(htmlDecode(tagsTree.name));
+                }
+            });
```
pingzing, line 56: If I wanted to _really_ pick some nits, we're setting the title here on every page change when we don't need to. But if someone pointed that out to me in my PR, I'd ignore it =P

pingzing (approved):

apps/bettafish/src/app/pages/tags/tag-page.component.ts
SigmasonicX: Actually no, that only happens if the tag ID in the URL doesn't match what it was before, that is to say if the user moves to a different tag.

pingzing: oh, duh, I even read that part too and went "ah yes, smart". ...then immediately forgot about it

PR #647, opened 2021-08-17
Feature/pseudonyms part 2

Figments:
work in progress


libs/api/database/src/lib/content/schemas/sections.schema.ts
```diff
@@ -1,14 +1,16 @@
...
-
 import { Section, AuthorsNotePos } from '@dragonfish/shared/models/sections';
 
 @Schema({ timestamps: true, autoIndex: true, collection: 'sections' })
 export class SectionsDocument extends Document implements Section {
     @Prop({ default: () => nanoid() })
     readonly _id: string;
 
+    @Prop({ trim: true, required: true })
+    readonly contentId: string;
```
SigmasonicX, line 12: What was the purpose of this change?

libs/api/database/src/lib/content/stores/browse.store.ts
```diff
@@ -74,10 +71,10 @@ export class BrowseStore {
         pageNum: number,
         kinds: ContentKind[],
         filter: ContentFilter,
-        userId?: string
+        userId?: string,
     ): Promise<PaginateResult<ContentDocument>> {
         const query = { kind: { $in: kinds }, 'audit.isDeleted': false, 'audit.published': PubStatus.Published };
-        const filteredQuery = await this.determineContentFilter(query, filter);
+        const filteredQuery = await BrowseStore.determineContentFilter(query, filter);
```
SigmasonicX, line 77: Why replace "this" with "BrowseStore" throughout this file and make the method static?

SigmasonicX (commented):
I have some questions, but otherwise, looks fine. I'll more thoroughly test it once the entire feature is in.

libs/api/database/src/lib/content/stores/browse.store.ts
Figments: recommendation of my IDE, to be honest

libs/api/database/src/lib/content/schemas/sections.schema.ts
Figments: sections currently don't have any association with the content they're part of, which makes it hard to figure out which section belongs to what work when those sections are removed from the work by the author. this change fixes that

SigmasonicX (approved):
LGTM

PR #654, opened 2021-08-26
Make invalid passwords not crash

pingzing:
## Description
We had a condition where, if a user passed in an invalid password, we'd return a null user (which the caller didn't check for) which meant we kept right on trucking along until exploding on the null. Now, we correctly throw an immediate 404.

(I also changed the logic to be a little easier to follow. Sentinel Pattern ftw)

## Areas Affected
- [x] Sign In/Out

### DIFF
--- a/apps/api/src/app/services/auth/auth.service.ts
+++ b/apps/api/src/app/services/auth/auth.service.ts
@@ -18,24 +18,25 @@ export class AuthService {
         private readonly accountStore: AccountsStore,
         private readonly pseudStore: PseudonymsStore,
         private readonly jwtService: JwtService,
-    ) {}
+    ) { }
 
     //#region ---ACCOUNTS---
 
-    public async validateAccount(email: string, password: string): Promise<Account> {
+    public async validateAccount(email: string, password: string): Promise<Account | null> {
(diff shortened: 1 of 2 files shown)


apps/api/src/app/services/auth/auth.service.ts
```diff
@@ -18,24 +18,25 @@ export class AuthService {
...
-            } catch (err) {
-                throw new InternalServerErrorException(`Something went wrong! Code: sunfish`);
+        if (!potentialAccount) {
+            throw new NotFoundException(`No user with those credentials found. Have you signed up?`);
+        }
+
+        try {
+            if (!(await verify(potentialAccount.password, password, { type: argon2id }))) {
+                this.logger.warn(`Someone attempted to log into Account ${potentialAccount._id} and failed!`);
+                throw new NotFoundException(`No user with those credentials found. Have you signed up?`);
```
SigmasonicX: I'm pretty sure this is just caught by the catch block. If we want to surface the NotFoundException and not the InternalServerErrorException, this will need rearranging.

SigmasonicX (changes requested):

SigmasonicX (approved):
LGTM

PR #657, opened 2021-08-27
Fixes tag link display on Dashboard and improves tag page header

SigmasonicX:
## Description
The tag management link on the Dashboard doesn't match other links on the page in appearance, and the tag page header is a bit hard to read.

## Changes
Changes the type of link used on the Dashboard to match the others, and creates a CSS file for tag-page to adjust the appearance of the header.

## Areas Affected
- [ ] Site Navigation
- [ ] Home
- [ ] Sign In/Out
- [ ] Registration
- [ ] Settings

.
- [ ] Profile
- [ ] My Stuff (main page)
- [ ] Editor
- [ ] Browse
- [ ] Work Card

.
- [ ] Work Page
- [ ] Blog Page
- [ ] Collections
- [ ] Comments
- [x] Other Pages

.
- [ ] Account Authentication
- [x] Dashboard
- [ ] Mobile

### DIFF
--- a/apps/bettafish/src/app/pages/tags/tag-page.component.scss
+++ b/apps/bettafish/src/app/pages/tags/tag-page.component.scss
@@ -1,3 +1,0 @@
-div.tag-header {
-    font-size: 24px;
-}
--- a/apps/bettafish/src/app/pages/tags/tag-page.component.scss
+++ b/apps/bettafish/src/app/pages/tags/tag-page.component.scss
@@ -1,0 +1,3 @@
+div.tag-header {
+    font-size: 24px;
+}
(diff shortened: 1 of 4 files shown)


apps/bettafish/src/app/pages/tags/tag-page.component.scss
```diff
@@ -0,0 +1,3 @@
+div.tag-header {
+    font-size: 24px;
```
pingzing: Is there not a tailwind class for this?
Also, it's better practice to use `em`s or `rem`s for text (though we break this rule _all the time_ and we shouldn't >.>)

pingzing (dismissed):

apps/bettafish/src/app/pages/tags/tag-page.component.scss
SigmasonicX: Ah, didn't think of Tailwind. Made that change.

pingzing (approved):

PR #658, opened 2021-08-28
Add various VSCode and Prettier and ESlint settings to make Dragonfish warn and autoformat

pingzing:
This doesn't touch any actual code files, but it enables a number of handy editor things in VSCode. Note that this only applies to the Dragonfish project, as I'm less comfortable imposing formatting rules on the frontend, as I'm less familiar with its idiosyncrasies. You could probably use this PR as a template though.

- It turns on auto-format and auto-fix on save by default. This will automatically format the file and apply any automatic fixes from Prettier or ESLint whenever you save.
- It turns on CodeLens by default, which is just handy. It displays a little inline annotation in the file that shows you how often public members (functions & variables) are references externally.
- It will warn on any function that doesn't have an explicit return type
- It will warn on any parameter that doesn't have an explicit type
- It will warn if you have an `async` function that doesn't use `await`.

IN ADDITION:
I also bumped Dragonfish's compile target from ECMAScript 2015 to ECMAScript 2019, as that's the highest that our runtime version of Node can safely support. This will unlock a small handful of new language-level improvements.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
         "angular.ng-template",
         "ms-vscode.vscode-typescript-tslint-plugin",
         "esbenp.prettier-vscode",
+        "dbaeumer.vscode-eslint",
         "firsttris.vscode-jest-runner"
     ]
 }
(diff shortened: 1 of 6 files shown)


SigmasonicX (approved):
LGTM

Figments (approved):

PR #663, opened 2021-08-30
Feature/pseudonyms part 4

Figments:
- Updates the Settings page by doing the following:
    - Moves Settings to its own lazily-loaded module for easier consolidation of components and views
    - Adds various different Categories for better settings organization
    - Enables the ability to edit any pseudonyms, and to switch between them
    - Temporarily disables account editing (e.g. email and password changes) pending the implementation of the mail system
    - Removes old components from the main bettafish directory


SigmasonicX (approved):
LGTM, but after setting a tagline, there's no way to remove it, since the minimum is three characters.

PR #667, opened 2021-09-05
feature/pseudonyms-part-5

Figments:
this PR removes portfolios and migrates them to profiles, encapsulated in their own module.


SigmasonicX (changes requested):
Code-wise, this seems fine, but the profile header and its animations don't look good on my laptop.

The animations for going between topbar modes looks bad.

The top is cut off
image

It takes up a lot of screen real estate when scrolling down
image

And we should consider how it looks on mobile
image

pingzing:
Gave the diff a quick once-over and I'm inclined to agree with Sigma--the actual changes look fine, but I really don't like the sticky header. Even if the cutoff and polish issues were fixed, sticky headers takes up _so_ much vertical real estate that's it's a really unpleasant experience. You can always just scroll back to the top--none of our pages are particularly long.

SigmasonicX (approved):
LGTM

PR #672, opened 2021-09-24
Implements searching by kind and pagination

SigmasonicX:
Addresses part of ticket #637

## Description
As part of more advanced searching, we need to distinguish searches by kind.

## Changes
- Implements searching by kind and pagination
- Adds field to select kind
- Removes original search method of showing results of three kinds
- Newly implements ability to search news

## Areas Affected
- [ ] Site Navigation
- [ ] Home
- [ ] Sign In/Out
- [ ] Registration
- [ ] Settings

.
- [ ] Profile
- [ ] My Stuff (main page)
- [ ] Editor
- [ ] Browse
- [ ] Work Card

.
- [ ] Work Page
- [ ] Blog Page
- [ ] Collections
- [ ] Comments
- [x] Other Pages

.
- [ ] Account Authentication
- [ ] Dashboard
- [ ] Mobile

### DIFF
--- a/apps/api/src/app/controllers/search/search.controller.ts
+++ b/apps/api/src/app/controllers/search/search.controller.ts
@@ -9,6 +9,7 @@ import { InitialResults } from '@dragonfish/shared/models/util';
 import { ContentFilter } from '@dragonfish/shared/models/works';
 import { DragonfishTags } from '@dragonfish/shared/models/util';
 import { ISearch } from '../../shared/search';
+import { SearchKind } from '@dragonfish/shared/models/search';
 
 @Controller('search')
 export class SearchController {
@@ -27,7 +28,7 @@ export class SearchController {
     @Get('find-related-content')
     async findRelatedContent(
         @Query('query') query: string,
-        @Query('kind') kind: string,
+        @Query('kind') kind: SearchKind,
         @Query('pageNum') pageNum: number,
         @Cookies('contentFilter') contentFilter: ContentFilter
     ): Promise<PaginateResult<ContentModel>> {
--- a/apps/api/src/app/controllers/search/search.controller.ts
+++ b/apps/api/src/app/controllers/search/search.controller.ts
@@ -24,6 +24,17 @@ export class SearchController {
     }
 
     @ApiTags(DragonfishTags.Search)
+    @Get('find-related-content')
+    async findRelatedContent(
+        @Query('query') query: string,
+        @Query('kind') kind: string,
+        @Query('pageNum') pageNum: number,
+        @Cookies('contentFilter') contentFilter: ContentFilter
+    ): Promise<PaginateResult<ContentModel>> {
+        return await this.searchService.findRelatedContent(query, kind, pageNum, contentFilter);
+    }
+
+    @ApiTags(DragonfishTags.Search)
     @Get('get-user-results')
     async getUserRequests(
         @Query('query') query: string,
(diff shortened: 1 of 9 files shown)


apps/api/src/app/controllers/search/search.controller.ts
```diff
@@ -23,6 +23,17 @@ export class SearchController {
         return await this.searchService.fetchInitialResults(query, contentFilter);
     }
 
+    @ApiTags(DragonfishTags.Search)
+    @Get('find-related-content')
+    async findRelatedContent(
+        @Query('query') query: string,
+        @Query('kind') kind: string,
```
pingzing: You should just be able to use the `SearchKind` enum here directly, and Nest will coerce it from a string into a SearchKind.

apps/api/src/app/services/search/search.service.ts
```diff
@@ -46,6 +47,45 @@ export class SearchService implements ISearch {
         return result;
     }
 
+    async findRelatedContent(
+        query: string,
+        searchKind: string,
```
pingzing: Ditto the enum here, then you won't have to do the `searchKind as SearchKind` below.

libs/shared/src/lib/models/search/search-kind.enum.ts
```diff
@@ -0,0 +1,8 @@
+export enum SearchKind {
```
pingzing, line 1: I'm looking at this and thinking that if, in the future, we want to allow searching by kinds as a series of checkboxes, rather than a single radio button or dropdown, this enum style approach won't work. I'd recommend that either:

a ) we use a [Flags style enum](https://basarat.gitbook.io/typescript/type-system/enums#number-enums-as-flags) here, and do some bitshifty stuff to figure out what's inside of it on the backend
b) We change this enum to just be 
```TypeScript
    Prose = "Prose",
    Poetry = "Poetry",
    Blog = "Blog",
    News = "News",
    User = "User",
```
...and pass up an array of all the options the user has selected, instead of a single enum. (You can do that with query params, You can see an example of how we do it the `ContentKind` enum over in [`fetch-all-new`](https://github.com/OffprintStudios/dragonfish/blob/3840d9f67092e0a2ec632e58d935e6509dad55ea/libs/client/services/src/lib/dragonfish-network.service.ts#L261) in the NetworkService. The controller that receives that is the BrowseController.)

apps/api/src/app/services/search/search.service.ts
```diff
@@ -46,6 +47,45 @@ export class SearchService implements ISearch {
...
+                kinds.push(ContentKind.PoetryContent);
+                break;
+            case SearchKind.ProseAndPoetry:
+                kinds.push(ContentKind.PoetryContent, ContentKind.ProseContent);
+                break;
+            case SearchKind.Prose:
+                kinds.push(ContentKind.ProseContent);
+                break;
+            case SearchKind.User:
+                // Doesn't handle, call searchUsers() instead
```
pingzing: I would make this throw an error, because there's no way to logically continue if that's what the user has passed to us. 400 BAD REQUEST, with an error message telling the user that they should call `/search/get-user-results` instead.

pingzing (changes requested):
I'd say a medium-sized change around dealing with SearchKind, but the core of this looks good. Great work!

libs/shared/src/lib/models/search/search-kind.enum.ts
SigmasonicX: I don't think having multiple selection is the right move. In particular, if you want to stop searching for prose and poetry and only search for users, then requires both unselecting prose and poetry and selecting users. If we display multiple kinds on the search results, that causes complications because they use different cards and, if they're displayed separately, then figuring out pagination could result in complications. I think it makes more sense to just display results for one kind (and Prose and Poetry), which also matches how it is on other sites.

pingzing (approved):

PR #720, opened 2021-10-15
Feature/notifications rework

Figments:
Initial progress on the rework of our notifications system. Changes include:

- moving notifications to a dedicated nestjs module
- replacing the mongo-backed custom queue system with `@nestjs/bull`
- simplifying how notifications are implemented
- creating skeleton files and schemas for future use
- changing the user menu on the client side to include notifications management

right now, users will only get notifications for comments on their content. over time, we will add other functionality like "when a chapter gets added to a work i favorited", following users, etc.


SigmasonicX (approved):
LGTM
In testing, I wasn't able to get notifications to work, so make sure others test this before it's pushed to live

PR #771, opened 2021-12-07
adding cors

Figments:
this PR enables CORS for any localhost address, any offprint subdomain, and offprint.net itself

### DIFF
--- a/libs/api/database/src/lib/comments/schemas/index.ts
+++ b/libs/api/database/src/lib/comments/schemas/index.ts
@@ -18,7 +18,7 @@ export { ContentCommentDocument, ContentCommentSchema } from './content-comment.
  * Sets up the comment collection.
  * @returns
  */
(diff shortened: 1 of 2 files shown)


libs/api/database/src/lib/comments/schemas/index.ts
```diff
@@ -18,7 +18,7 @@ export { ContentCommentDocument, ContentCommentSchema } from './content-comment.
  * Sets up the comment collection.
  * @returns
  */
-export async function setupCommentCollection(): Promise<Schema<CommentDocument>> {
+export async function setupCommentCollection() {
```
pingzing, line 21: figs, what. stop getting rid of return type annotations >:

pingzing (changes requested):

pingzing (approved):

PR #808, opened 2022-04-01
Update SelectProfile.svelte

Figments:
fixed an issue where profiles were being set incorrectly whenever a user tried to change things around


SigmasonicX (approved):
LGTM

PR #809, opened 2022-04-02
fixed issue preventing filter settings being saved

Figments:
filter settings should now save as expected. i also separated out each of the general settings boxes into their own components for better maintainability


SigmasonicX (approved):
LGTM

PR #828, opened 2022-04-04
Enhancement/remove autopopulate from content

Figments:
removes autopopulate from content fetching for increased performance. should also hopefully help with failing notifications, since i believe jobs were hanging on the fetch requests.


SigmasonicX (changes requested):
In the Approval Queue, after I claim a work, the claimed work's title disappears and I can't view it.
`Uncaught (in promise) TypeError: ctx[6].workToApprove.author is undefined`

SigmasonicX (approved):
LGTM now

PR #835, opened 2022-04-08
Enhancement/section loading

Figments:
- separates out content, section, ratings, and library doc loading into four distinct requests
- decouple sections from client-side state management
- promote use of disabling requests and mutations depending on client auth state

this should overall increase request and load performance of content pages overall, as well as make it easier to manage certain features based on current auth state.


SigmasonicX (approved):
LGTM
Also fixes #836

PR #841, opened 2022-04-11
feature/followers-rework

Figments:


apps/api/src/modules/content/db/stores/content.store.ts
```diff
@@ -366,22 +366,54 @@ export class ContentStore {
...
+                }
+            }
+
+            sec.published = pubStatus.newPub;
+
+            if (pubStatus.newPub === true) {
+                sec.audit.publishedOn = new Date();
+            } else {
+                sec.audit.publishedOn = new Date();
+            }
```
SigmasonicX, line 400: Either there was a mistake with this clause or there shouldn't be an if check.

SigmasonicX (approved):
Testing, the side menu doesn't update with the number of followed users / followers until you refresh. Otherwise, looks fine.

PR #873, opened 2022-06-27
Fix/content not loading

Figments:
Addresses the following issues:

- updates dependencies for sveltekit and svelte-select
- fixes content links not loading when navigated to directly

Additionally:

- updates CORS policy to include API access for preview URLs on cloudflare
- updates minimum node version to 16.10.0 in accordance with SvelteKit and NestJS minimums
- updates README to reflect node version changes


SigmasonicX (approved):
LGTM

